PR #14, opened 2019-02-21
Fixed asc to a previous hash

willemneal:
- Added a debug function to call out to debugger.
- Added source map to output so that wasm functions have names.


jtenner:
LGTM! I will pull in an hour or so after a more detailed review.

PR #45, opened 2019-05-10
Added files flag

willemneal:
--file or -f now filters out tests file names via cli


jtenner:
FWIW the lib conflicts can be completely ignored, and the package-lock can be deleted/regenerated later.

willemneal:
@jtenner Not sure why the coverage dropped, but otherwise how does it look?

jtenner:
Idk why coverage dropped either. 

Among other things It looks like the types in the package.json need to point to lib/as-pect.d.ts

jtenner:
I promise to take a look at this tomorrow. I just happen to be very busy today.

jtenner:
The file affected looks like src/test/TestResult.ts:102

This is for variance calculations. Is there any way to force a pull request that doesn't pass onto master?

jtenner:
I will handle this pull request tomorrow for sure. Sorry I couldn't get around to it. Mother's day and all!

PR #48, opened 2019-05-17
Refactor cli v.0.6.1

jtenner:
This pull request refactors the functions out into their own separate modules.

Later changes to functionality will be deferred for v 0.7.0


src/cli/cli.ts
```diff
@@ -0,0 +1,55 @@
+
```
willemneal: My one complaint is why is this not `index.ts`?  Then you wouldn't have to change the import path other places.

src/cli/run.ts
```diff
@@ -0,0 +1,275 @@
...
+    "--debug": [],
+    "--measure": [],
+    "--sourceMap":[],
+    /** This is required. Do not change this. */
+    "--binaryFile": ["output.wasm"],
+  };
+  const disclude: RegExp[] = configuration.disclude || [];
+  let reporter: TestReporter = configuration.reporter || new DefaultTestReporter();
+
+  const performanceConfiguration = configuration.performance || createDefaultPerformanceConfiguration();
```
willemneal: One other idea is to break this out as it's own class which you pass yargs in the constructor.  Then the getters would handle the default case or read from yargs.

src/cli/cli.ts
jtenner: This sounds like good criticism. I'll make the change.

src/cli/run.ts
jtenner: I don't understand what you mean?

willemneal:
```ts
class PerformanceConfiguration {
    constructor(private args: yargs){}

    get maxSamples(): string {
        return this.args.argv.maxSamples ? parseFloat(this.args.argv.maxSamples.toString()) : null;
    }
}
```

jtenner: I will refactor this section in the next pull request. It's a breaking change and should get a major version bump.

PR #49, opened 2019-05-17
Refactor cli

jtenner:
There's a bunch of things here.

- [x] Readme clarifications
- [x] Make `run.ts` smaller by pulling out small functions
- [x] use `npm-run-all` to parallelize the typescript build
- [x] change `test.ts` to match `bin/asp`


package.json
```diff
@@ -1,12 +1,16 @@
...
   "description": "Write AssemblyScript tests at blazing fast speeds",
   "main": "lib/index.js",
   "types": "./assembly/__tests__/as-pect.d.ts",
   "scripts": {
-    "test": "npm run tsc && node lib/test && jest",
-    "tsc": "tsc && tsc --outfile lib/as-pect.amd.js --module amd",
+    "test": "run-s tsc asp jest",
+    "asp": "node lib/test",
+    "jest": "jest",
+    "tsc": "run-p tsc:default tsc:amd",
```
willemneal: Nice!  Didn't know about that feature!

src/cli/util/collectPerformanceConfiguration.ts
```diff
@@ -0,0 +1,28 @@
+import { IYargs } from "./IYargs";
+import { IPerformanceConfiguration } from "../../util/IPerformanceConfiguration";
+
+export function collectPerformanceConfiguration(yargs: IYargs, performanceConfiguration: IPerformanceConfiguration) {
+  if (yargs.argv.hasOwnProperty("performance"))
```
willemneal: If this is false shouldn't you return from the function?  If you aren't going to collect performance data than why check every flag?

package.json
jtenner: Yeah, it's a great feature. I use it for setting up development servers in complicated setups.

src/cli/util/collectPerformanceConfiguration.ts
jtenner: The `--performance` flag enables performance statistics gathering on *EVERY* test. This flag may not be present and it's still possible that `--collect-variance`, (for instance,) might be present. This should override what is currently in the `as-pect.config.js` file.

willemneal: Ah okay.  I forgot about the programmatic way of recording stats.

PR #53, opened 2019-05-22
[Fix npmignore] comments and remove test configurations

jtenner:
fixes #52

Issue #29, opened 2019-04-22
[Feature] Warn users when tests are configured improperly


jtenner:

jtenner:
We need to tackle this next.

jtenner:
I have put this off long enough! It's time to emit some warnings!

PR #62, opened 2019-05-24
Runtime and Many New Features Support

### LINKED ISSUE #6: Default Reporter should pipe log values to stdout instead of console.log()

### LINKED ISSUE #16: Add toInclude and toIncludeEqual

### LINKED ISSUE #52: [Bug] .npmignore not including certain files
# Files Not Properly Ignored by `.npmignore`

- `__tests__`
- `assembly/__tests__` (don't forget to *include* the types file)
- `jest.config.js`

### LINKED ISSUE #56: Feature requests: filtering tests by name
It would be nice if `as-pect` can filter tests by name like Rust, although since tests usually only have description, we might want to consider some other way of filtering.

### LINKED ISSUE #58: [Feature] --no-run and --output-binary flag
We need a way to capture the output of the compiler so that binaries can be saved and run later, say, from a web server.

### LINKED ISSUE #60: [Feature] Support Compiler Flags From Command Line

### LINKED ISSUE #61: [Feature] Support AssemblyScript Runtime
This is already in progress. There are some issues with the way references are compared and they need to be refractored.

### LINKED ISSUE #63: [Feature] CSV/JSON reporter
This is good for test reports.  For example, when grading students it would be handy to produce a CSV with each group summarized with number of passing tests. e.g.

```
2,1,10
```

### LINKED ISSUE #73: Reporting Expected and Actual `u64` and `i64` values doesn't work
Any advice on the matter @willemneal ?

jtenner:
closes #61 
fixes #53 
fixes #52 
closes #56 
closes #58 
closes #16 
closes #60 
closes #64 
closes #63 
closes #6 
fixes #73


assembly/__tests__/setup/Vec3.ts
```diff
@@ -13,31 +13,27 @@ export class Vec3 {
...
 
-  @inline
   @operator("<")
   protected __lessThan(reference: Vec3): bool {
     var magnitude = this.magnitude();
     var refmagnitude = reference.magnitude();
     return magnitude < refmagnitude;
   }
 
-  @inline
```
willemneal: Just curious why this change?

jtenner:
The assemblyscript compiler emits a warning when it tries to inline an
operator like this. When it becomes supported I will add it back.

assembly/__tests__/setup/Vec3.ts
jtenner: The assemblyscript compiler emits a warning when it tries to inline an
operator like this. When it becomes supported I will add it back.

jtenner:
The branch now no longer has 100% TestContext line coverage.

We will need to test the `testRegex` and `groupRegex` matchers.  This is very easy, but requires an additional binary compilation when jest runs. Will work on this tomorrow.

jtenner:
Coverage is now at 100%. Halfway finished with `--output-binary` and `--no-run`. While, `--output-binary` is completed, `--no-run` requires that no output is pushed to the console about the tests. Will need some more time to figure out how this flag works.

jtenner:
The `--norun` flag is now added. `--no-run` is parsed by `yargs` as `run: false` instead of `noRun: true`.

jtenner:
The next feature that needs to be added is compiler flags from the command line. 

This is going to require a huge cli parser refactor and should probably ship with `2.0.0`. Do you have any objections @willemneal. This can include #60  and #29.

If this is the case, all we need to do is add a new assertion called `toInclude` and `toIncludeEqual` in the assertions.

```ts
const array: number[] = [1, 2, 3];
expect(array).toInclude(2, "array should have 2");
```

willemneal:
Yeah I think it's fine to wait on the command line args.  Though I don't think it'll require as much rewriting as you suggest.  The way that npm lets you pass arguments to scripts is by using `--`  So we just have to split over the ` -- ` and then pass the first half to yargs.

jtenner:
Okay. I added toInclude and will implement  toIncludeEqual.

If you want, @willemneal you can implement command line args.

jtenner:
The only problem I see with that is that command line args should overwrite compiler flags passed via configuration

jtenner:
Is there any good way to override compiler flags provided by the configuration?

assembly/__tests__/setup/Vec3.ts
jtenner: All set? I don't want to mark this conversation as resolved without running it by you first.

willemneal:
I'll make a fork of the compiler flags and add the overwritting.  I think the easiest way is to use Object.assign, which will use the compiler flags passed first and the config file flags won't overwrite them.

jtenner:
Yeah there's only one problem. There are some flags you can pass multiple times. Is there a way to handle these?

src/test/TestContext.ts
```diff
@@ -1,139 +1,100 @@
...
+    this.startupTime = timeDifference(performance.now(), start);
+
+    if (this.errors.length > 0) return;
+    this.ready = true;
 
     // start the test suite
     this.reporter.onStart(this);
-    runContext.start = performance.now();
 
     testgroup:
```
willemneal: This is never referenced.

willemneal:
@jtenner Since we're using the same argument parser as asc, then it will be same as if they passed multiple arguments manually to `asc.`  Not sure what would happen, but that's outside the scope of this project.

src/test/TestContext.ts
jtenner: What's not referenced? The `start` and `end` values used by `performance.now()` aren't stored on the `TestContext`.

It's just used for calculating the `.time` value.

jtenner: I removed this label

jtenner:
I'm glad I took the time to generate this pull request. It contains a lot of bug fixes and features that I think are necessary to make `as-pect` enterprise-grade testing software.

I think I need to spend some more time with the documentation in order to make this pull request more complete.

jtenner:
I think this is now *very* stable, but I would love to release this as a beta. This is a backward incompatible rewrite because of all the memory layout changes, so it will be released as `1.0.0-beta`.

@willemneal I think all this source code could use a huge code review when you get the chance. There are lots to optimize and things to comment. Please help me point them out!

willemneal:
Okay great!  I just tested this with my new compiler frontend and am getting a runtime of 4.2 s.  Should I make a PR to this branch to add my assemblyscript as the dependency and change the way it's imported.

jtenner:
Yes please. Make a fork so I can review it. I want to make sure I understand what you did.

jtenner:
I need to make one more commit.

assembly/__tests__/toInclude.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { Vec3 } from "./setup/Vec3";
+
+let numberTester: i32[] = new Array<i32>();
```
willemneal: Shouldn't this be in a `beforeAll` block?

assembly/internal/comparison/arrayComparison.ts
```diff
@@ -0,0 +1,44 @@
+import { assert } from "./assert";
+import { reportActual } from "../report/reportActual";
+import { reportExpected } from "../report/reportExpected";
+
+// @ts-ignore: Decorators *are* valid here
+@inline
+export function arrayComparison<T>(actual: T, expected: T, negated: i32, message: string): void {
```
willemneal: Is `<T extends Array>` possible?  If not it would be a good feature request for upstream.

assembly/internal/comparison/closeToComparison.ts
```diff
@@ -0,0 +1,34 @@
+import { reportActual } from "../report/reportActual";
+import { reportExpected } from "../report/reportExpected";
+import { assert } from "./assert";
+
+// @ts-ignore: Decorators *are* valid here!
+@inline
+export function closeToComparison<T>(actual: T, expected: T, negated: i32, decimalPlaces: i32, message: string): void {
```
willemneal: Again it would be awesome if we could do `<T extends float>`, where float is `type float = f32 | f64;`  But again this feature should be pushed upstream.

jtenner: I like the idea that this assertion is a runtime error.

```ts
assert(isFloat<T>(), "Message");
```

This is good enough for now.

assembly/internal/comparison/arrayComparison.ts
jtenner: Yes. This is not possible. `Array` has a generic `T` type parameter that must be provided, and we cannot know what `U` is up front. This is why `Array<U>` where `U` is a reference type is not supported.

assembly/__tests__/toInclude.spec.ts
jtenner: This is a global value, and the value is created once.

If I want to add a future `describe()` block, it will add those items each time a `describe()` block is run.

I only want it to run once.

assembly/internal/report/reportExpected.ts
```diff
@@ -0,0 +1,171 @@
+import { ValueType } from "./ValueType";
```
willemneal: The reporter files are certainly improved, but still feels like there is code duplication going one.  Couldn't they be combined like `reportFloat(negated: i32, expect: f64, actual: f64)`.  If separating them makes sense then ignore me.

assembly/internal/comparison/closeToComparison.ts
willemneal: True, just saying that we can reduce runtime checks that should be compile time errors.

assembly/internal/report/reportExpected.ts
jtenner: This is because reporting a `float`, `i32`, or `u32` doesn't require `long` support. Reporting an `i64` or a `u64` is a special case that requires some JavaScript hackery.

willemneal: I meant in the more general case; having a single report function versus having two.

jtenner: Also, it uses the same external function and simply provides a different signature for convenience. Generic `T` types might be supported by AssemblyScript for linked functions like this, but this makes things very clear about what is going on.

Just about the only argument to be made for using the "same" function with a generic parameter is speed of execution, which is negligable, since the compiler would need to compile two versions of the function anyway.

assembly/internal/report/reportExpected.ts
```diff
@@ -0,0 +1,171 @@
...
+@external("__aspect", "reportExpectedValue")
+declare function reportExpectedInteger(value: i32, negated: i32): void;
+
+// @ts-ignore: Decorators *are* valid here!
+@external("__aspect", "reportExpectedReference")
+declare function reportExpectedReferenceExternal(value: usize, offset: i32, negated: i32): void;
+
+// @ts-ignore: Decorators *are* valid here!
+@external("__aspect", "reportExpectedString")
+declare function reportExpectedString(value: string, negated: i32): void;
```
willemneal: What I'm trying to say is why not have `reportString(expect: string, actual: string, negated: i32): void`, etc, instead of breaking it into two calls?

assembly/internal/comparison/closeToComparison.ts
jtenner: This is incorrect. We don't know what `T` is at compile time until the function is called. 

`T` itself is a parameter of the `expect<T>()` function, thus, we cannot assert that `T` is a float without a runtime check.

assembly/internal/report/reportExpected.ts
jtenner: I understand now.

This was a personal development choice I made because sometimes we compare a float value to a string (a la `toBeFinite()` which compares against the string "finite".) Leaving things separate makes things clearer and does not make things too much slower. Remember that test files of reasonable size run in `25ms`. This has "premature" optimization written all over it.

assembly/internal/comparison/closeToComparison.ts
willemneal: Reading this I thought that `expect<T>(x).toBeCloseTo` means that the type parameter would be passed to closeToComparison<T>.  Thus at compile time the type could be checked to make sure it extends float (that is if that is possible in the first place).

```ts
 @inline
  public toBeCloseTo(expected: T, decimalPlaces: i32 = 2, message: string = ""): void {
    closeToComparison<T>(this.actual, expected, this._not, decimalPlaces, message);
  }
```

jtenner: We don't know if `T` is an `f32` or an `f64` at compile time. We can only pass the type along to the comparison.

willemneal:
What is reasonable size?  And regardless all I'm saying is that these seem
like duplicated code. Which makes it feel like something is not right. I'm
fine as is just saying easy to develop at first doesn't always mean what it
should always be.

assembly/internal/comparison/closeToComparison.ts
jtenner: Also, that function should not be inlined.

assembly/internal/report/reportExpected.ts
jtenner: Modules of reasonable size include a test file of at least 50 tests.

For instance the `toStrictEqual()` tests run in less than `8ms` despite the fact that there are 25 tests. Asking for more performance out of these functions will produce a very minute performance gain, especially since they are only called when a test fails.

assembly/internal/comparison/closeToComparison.ts
willemneal:
```
  it("should expect values that are close", (): void => {
    expect<f64>(1.0).toBeCloseTo(1.00001, 2, "1.000001 is close to 1.0");
  });
```
The compiler does know what `T` is.  My point was if we had a way to type check that the generic type was either floats then it would save the dynamic check.  It's my complaints with AS, not your code.

assembly/internal/report/reportExpected.ts
willemneal: For me it's not the efficiency, but rather the design choice, which like I said if it was put in place for a reason and changing would make things harder to work with don't bother.  Rather trying to understand the design choice of two files that seem almost identical.

willemneal:
LGTM!  Just my question about design, but those are just for my own understanding and gripes about the current state of assembyscript.

assembly/internal/report/reportExpected.ts
jtenner: I agree the files are identical and it feels like the code is repeated. I could be convinced to combine these functions together, but these classes are essentially static global variables that perform two slightly different explicit operations. If you can help me design something that is much clearer and makes things easy to work with, I will change it. Definitely.

In the meantime, this is the best compromise I could personally come up with.

jtenner:
@willemneal your experience is really valuable and I appreciate all your effort. When you think this is ready, I will begin the process of documenting each function, and its purpose.

assembly/internal/comparison/closeToComparison.ts
jtenner: Agreed.  I don't think there's an option to make this check happen at compile time.

jtenner:
@willemneal I increased the documentation by a lot and fixed signed integer logging and reporting. I thought it was weird when I saw `4294967294` get reported as `-2`.

One last change needs to be made because the `@start` decorator is going to be removed in favor of the `--explicitStart` flag.

jtenner:
@willemneal Can you approve the request? I did touch almost every single file in the project now.  Now that assemblyscript master has been updated, I'd like to finish this pull request.

PR #68, opened 2019-06-01
Wasm runner

jtenner:
Implement runtime support with wasm runner.

`src/test/TestContext#testPass()` throws an error saying an item doesn't exist in an array.

Will need to inspect further.


jtenner:
for now, closing in favor of #62

PR #74, opened 2019-06-05
Updated to use my compiler

willemneal:


jtenner:
Thanks for the pull request. I wanted to see how your compiler worked. Looks like the interface is exactly the same! Hopefully this change makes it to the assemblyscript master branch. In the meantime I'll keep this pull request as reference

PR #75, opened 2019-06-06
Use compiler that caches parsed files

willemneal:


jtenner:
This is great. Currently, I'm waiting for a pull request I just wrote to land on assemblyscript#master.
I can definitely accept this on the condition that you promise to keep that branch up to date with all the features of the `master` assemblyscript branch.

willemneal:
Definitely!

willemneal:
According to travis:

The last PR: ` Ran for 2 min 36 sec `
This one: ` Ran for 1 min 16 sec `

jtenner:
Do we have a way to default to *your* compiler, but also allow the developer to specify a compiler entry point if they have to use an older version?

I would love to see a `--compiler` flag and configuration option. (Not everyone will be using the master branch of assemblyscript like the folks at near protocol.)

willemneal:
I think this should be pulled out into a separate project.  An assemblyscript version manager, `avm`.  That way we depend on that one, which will default to my compiler, but allow arguments to use another.

jtenner:
Hey. I can still use this as a `devDependency` if you think you can pull latest changes in AssemblyScript#master and apply the caching changes to it.

At the very least, it would speed up `as-pect` internal tests. 👍 

These benefits cannot be transfered down to our users because we shouldn't assume what compiler they are using.

jtenner:
We are letting our users pick their compiler. This is probably not necessary.

PR #76, opened 2019-06-06
Update expectation assertions to avoid conditional compilation bugs

jtenner:


jtenner:
@willemneal this pull request was created be jedisct1 who notified me that a `toStrictEqual()` assertion was failing. The conditional compilation did not account for `T` where `T` is a value type instead of a reference correctly.

Fastly has confirmed that it works.

PR #95, opened 2019-06-16
Remove yargs -> V 1.3.0

### LINKED ISSUE #88: Test filtering doesn't work
This is what happens when I try to run use `-t`:
```
~/c/assemblyscript-json  ‹new_runtime*› $ npx asp -f round -t int32
       ___   _____                       __
      /   | / ___/      ____  ___  _____/ /_
     / /| | \__ \______/ __ \/ _ \/ ___/ __/
    / ___ |___/ /_____/ /_/ /  __/ /__/ /_
   /_/  |_/____/     / .___/\___/\___/\__/
                    /_/

  ⚡AS-pect⚡ Test suite runner [1.1.0]


[Log] Initializing types.
```

### LINKED ISSUE #92: Failed expectation in test with no expectations
I've got a test like this:

```ts
	describe('.traverse/.traverseVisible/.traverseAncestors', () => {
		test('they traverse up and down the tree', () => {
			a.name = 'parent'
			b.name = 'child'
			c.name = 'childchild 1'
			c.visible = false
			d.name = 'childchild 2'

			b.add(c)
			b.add(d)
			a.add(b)

			a.traverse(function(obj) {
				names.push(obj.name)
			})

		})
	})
```

It fails on `runtime`, `rtrace` and `configuration-warnings`, but there's not really any output on rtrace or config branches.

The only somewhat interesting output is on the runtime branch:

```
❌ src/as/glas/core/Object3D.spec.ts Pass: 1 / 2 Todo: 0 Time: 3.515ms
  Object3D.traverse/.traverseVisible/.traverseAncestors
    ❌ they traverse up and down the tree - 
      [Expected]: null
      [Actual]  : null
```

The odd part is that if you delete the `traverse` block, so:

```ts
	describe('.traverse/.traverseVisible/.traverseAncestors', () => {
		test('they traverse up and down the tree', () => {
			a.name = 'parent'
			b.name = 'child'
			c.name = 'childchild 1'
			c.visible = false
			d.name = 'childchild 2'

			b.add(c)
			b.add(d)
			a.add(b)
		})
	})
```

then the test passes.

### LINKED ISSUE #96: [Enhancement] Refactor TestGroup Collection
This should refactor the way test groups are collected to speed up test group collection.

We don't need to slice the function pointers at all, just have to change the way `TestContext#runBeforeAll` works to become recursive.

```ts
class TestGroup {
  fork(): TestGroup {
    const group = new TestGroup();
    group.parent = this;
    return group;
  }
}
```

### LINKED ISSUE #101: [Bug] Error: not a string: 0
Calling `env.abort()` with a null string causes `__getString()` to throw an error because `null` is not a string.

This prevents abort from doing its job.

`TestCollector#abort` is the problem. It should report "No Error message provided."

jtenner:
- [x] Remove Yargs, create custom parser
- [x] Lazy Requires with named exports
- [x] Set asc to devDependencies
- [x] Add `--compiler=./path` configuration flag relative to `process.cwd()`
- [x] Add `compiler` configuration which accepts either a path or a compiler via `require("assemblyscript/dist/asc")`
- [x] Clean up cli `[Log]` output to be more consistent
- [x] refactor TestGroup collection
- [x] Optimize when dependencies are required
- [x] Fix getString `null` pointer exceptions
- [x] Use snapshots for TestContext output
- [x] Fix buffer length assertion
- [x] Bump version to 1.3.0

Notes:
- We can work on this together and submit pull requests to the `remove-yargs` branch
- Don't switch to anything but `AssemblyScript` master branches (Let's make sure it hits master first)
- Switch back to named exports and lazy require each reporter as needed
- Try to keep code style consistent with my style (I may do some cleanups afterwards)
- Backwards compatibility is a first priority

Fixes #96 
Fixes #88 
Fixes #101 
fixes #92


jtenner:
I just have to do some more cleanups. Can you take a look again at this whole pull request and approve it if it's good?

@willemneal

lib/util/IConfiguration.d.ts
```diff
@@ -35,7 +36,7 @@ export interface IConfiguration {
     /**
      * Set the default performance measurement values.
      */
-    performance?: IPerformanceConfiguration;
+    performance?: Partial<IPerformanceConfiguration>;
```
willemneal: I have a broader question of why we need individual `.d.ts` files if this is already covered in the big `amd.d.ts` file?

src/cli/run.ts
```diff
@@ -82,49 +120,56 @@ export function run(yargs: IYargs, compilerArgs: string[]): void {
...
+      }
+    })
+  }
+  const performanceConfiguration = yargs.performance;
 
   // include all the file globs
   console.log(chalk`{bgWhite.black [Log]} Including files: ${include.join(", ")}`);
 
   // Create the test and group matchers
   if (!configuration.testRegex) {
```
willemneal: It seems that this and the `groupRegex` below are always false because we don't hint in the docs or in the config that you can place either in the config.  Furthermore, since the config defines included/excluded, I don't see a reason to have this feature.  If you never want a group or test to run don't write it!

lib/util/IConfiguration.d.ts
jtenner: I wanted the amd bundle and the amd.d.ts file to be included in the browser for test bootstrapping using the monaco editor.  Each individual types file is for tooling. Since AssemblyScript and TypeScript need to resolve types seperately, I thought it would be easier to play to vscode's strength and leverage the typescript compiler to generate each types file.  It was a personal choice and I probably should keep it this way at this point.

src/cli/run.ts
jtenner: I thought it would be possible for people to provide multiple `as-pect` configurations and including the groupRegex and testRegex was easy. 

You are correct. If you don't want the test to be tested, don't write it. At the same time, let's remain backwards compatible.

lib/util/IConfiguration.d.ts
willemneal: But I'm confused why you have the `types` field in the package.json as "./assembly/__tests__/as-pect.d.ts", when users are expected to never import from the package and have a local copy of this file in the `assembly/__test__` folder.  If you changed this to `./lib/as-pect.amd.d.ts` then the tooling would use it and you wouldn't need the individual ones.

src/cli/run.ts
jtenner: I've decided to go with your pull request. :)

lib/util/IConfiguration.d.ts
jtenner: Right. The types should be for typescript developers.

Please change the types field to point to the correct place.

jtenner: Just need to fix this types problem and we are golden.

jtenner:
The buffer length assertion will be fixed in the next compiler release and won't cause problems.

All set for pulling @willemneal

PR #104, opened 2019-06-18
Massive Test Refactor

jtenner:
Use test snapshot with names: `${group.name}~${test.name} ${propertyName}` on majority of test properties to verify output.

For now, if something is wrong it will be hard to look for. However, the test is more maintainable and it's easy to find the problem in the `.snap` file.

Any further property definition changes will definitely be caught by the jest suites.


jtenner:
```
Test Suites: 8 passed, 8 total
Tests:       22 passed, 22 total
Snapshots:   707 passed, 707 total
Time:        17.257s
Ran all test suites.
```

I am much prouder of this.

PR #106, opened 2019-06-19
Update readme.md

trusktr:


README.md
```diff
@@ -379,7 +382,7 @@ describe("vectors", () => {
 
   // this runs after each test function, and must be placed within the describe function
   afterEach(() => {
-    vec = null; // free the vector
+    vec = null;
```
trusktr: We probably don't even need the `vec = null` now, right?

jtenner: Yes. To free the vector, because it's a global, we must set the global to  `null`.

jtenner:
LGTM! Thanks for your help. If you can find more stuff, I'll be happy to pull the changes.

PR #122, opened 2019-07-09
[Docs] small fixes

jtenner:
I noticed a few small problems with the docs. @willemneal can you approve this quickly?

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,7 +586,7 @@ Regardless of the installed version, all the compiler flags will be passed to
 the `asc` command line tool.
 
 ```ts
-const compiler = require(require.resolve(path.join(options.compiler, "cli/asc"));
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -586,7 +586,7 @@ Regardless of the installed version, all the compiler flags will be passed to
 the `asc` command line tool.
 
 ```ts
-import asc from "assemblyscript/cli/asc";
+const compiler = require(require.resolve(path.join(options.compiler, "cli/asc"));
```
willemneal: Isn't it `path.resolve`?

jtenner:
Right. Actually it's path.join(cwd(), compiler)

README.md
jtenner: Yeah it's path.join(cwd(), options.compiler)

jtenner: This should be all set.

PR #123, opened 2019-07-10
Switch to Mono Repo V2.1.0

willemneal:
Add:
- [x] `aspect/core`
- [x] `aspect/cli`
- [x] `aspect/docs` -> output to `./docs/*.*`
- [x] switch to typedocs
- [x] fix docs output (some variables are listed when they shouldn't be)


jtenner:
# TODO

1. Move most tests to `core`
2. Move tests that involve cli testing (using jest only) to `@as-pect/cli`
3. Setup typedocs to use `./packages/**/src/*.ts` and then add markdown files in `@as-pect/docs`

jtenner:
We need to move the comparison functions from the core module to the assembly module

willemneal:
LGTM!

PR #134, opened 2019-07-18
Version 2.2.1

jtenner:


jtenner:
@willemneal This pull request is completed. It adds some new features and uses static checks to throw compile-time errors.

Please review.

Thanks!

jtenner:
Because of the compiler change that needed to be made to become compatible with the latest update, this should be marked as `2.3.0`.

@willemneal I'm going to skip the review process this time around because brand new projects created with `as-pect` are broken without this update.

You can definitely submit a review for this and help me work on `2.3.1` if there are any problems. Thanks for understanding.

PR #142, opened 2019-08-19
Allow files with `.ts` to bypass current writeFile's behavior

willemneal:
Also now tries to import `assemblyscript/cli/asc`, which makes testing compiler changes easier and not have to have a webpack build.


packages/cli/src/run.ts
```diff
@@ -70,14 +70,24 @@ export function run(cliOptions: Options, compilerArgs: string[]): void {
   let parse: any;
 
   try {
```
jtenner: Might this be better?

```ts
for (const folder of ["cli", "dist"]) {
  try {
    asc = require(path.join(assemblyScriptFolder, folder, "asc"));
  } catch (ex) {

  }
}
```

jtenner: any thoughts? I think I intend to pull the request, but that might be a little easier to understand.

PR #153, opened 2019-09-10
Version 2.4.2

### LINKED ISSUE #148: [Fix] inconsistent init output

jtenner:
Fixes #148

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,9 @@ class Vec3 {
 
   // override the operator
   @operator("==")
-  protected __equals(ref: Vec3): bool {
+  protected __equals(ref: Vec3 | null): bool {
+    if (this === ref) return true;
+    if (ref === null) return false;
     return this.a == ref.a && this.b == ref.b && this.c == ref.c;
   }
 }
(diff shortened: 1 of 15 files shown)


lerna-debug.log
```diff
@@ -1,2 +0,0 @@
-30 error TypeError: Cannot set property 'committish' of undefined
```
willemneal: Is this class needed?

jtenner: Hang on. What is this?

README.md
```diff
@@ -177,7 +177,9 @@ class Vec3 {
 
   // override the operator
   @operator("==")
-  protected __equals(ref: Vec3): bool {
+  protected __equals(ref: Vec3 | null): bool {
+    if (this === ref) return true;
```
willemneal: Is `===` explained anywhere?  Otherwise should we or switch to converting to usize to avoid explaining it?

jtenner: How do you think this should be handled?

`changetype<usize>(this) == changetype<usize>(ref)` is definitely more complicated and harder to understand.

willemneal: Yeah perhaps just a comment will be enough.

jtenner:
```
// This is an exact pointer comparison
```

PR #181, opened 2020-01-25
Refactor RTrace & provide more utility functions

9oelM:
- [x] Refactor RTrace & provide more utility functions
- [x] Update `README.md` accordingly
- [x] Update tests accordingly (make tests pass & add more tests)


packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -1,3 +1,6 @@
+import { BLOCK, BLOCK_OVERHEAD } from "assemblyscript/std/rt/common";
```
jtenner: Should just be `"rt/common"` and `"rt/pure"`

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -220,23 +223,49 @@ export class RTrace {
...
+   * @returns {u32} - The type id of the allocated block.
+   */
+  public static typeIdOfReference<T>(reference: T): u32 {
+    if (!isReference<T>()) ERROR("Cannot get Type ID when T is not a reference.");
+    if (isFunction<T>()) ERROR("Cannot get Type ID of function reference.");
+    if (isNullable<T>()) {
+      assert(reference != null, "Cannot get Type ID of reference that is null.");
+    }
+
+    return this.typeIdOf(changetype<usize>(reference));
```
jtenner: `RTrace.typeIdOf(...);`

jtenner: We can't use `this` because these functions should be `static`.

packages/core/README.md
```diff
@@ -553,6 +553,16 @@ afterEach(() => {
...
 
+### RTrace.typeIdOf(pointer: usize)
+
+### RTrace.typeIdOfReference<T>(reference: T)
+
+### RTrace.sizeOf(pointer: usize)
+
+### RTrace.sizeOfReference<T>(reference: T)
+
+### RTrace.refCountOf(ptr: usize)
```
jtenner: Don't forget `RTrace.refCountOfReference<T>(reference: T)`

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -271,7 +300,7 @@ export class RTrace {
    * └─┴─────┴───────────────────────────────────────────────────────┘
    */
   public static refCountOf(ptr: usize): u32 {
```
jtenner: When adding `refCountOfReference<T>(reference)` we may need to subtract `1` because the refCount at the point of measurement should be `+1` in that function. This is because calling `refCountOfReference` itself will increment the count by design.

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -220,23 +223,49 @@ export class RTrace {
...
+  }
+
+  /**
+   * Get the type id (class id) of a reference.
+   *
+   * @param {T} reference - The reference.
+   * @returns {u32} - The type id of the allocated block.
+   */
+  public static typeIdOfReference<T>(reference: T): u32 {
+    if (!isReference<T>()) ERROR("Cannot get Type ID when T is not a reference.");
```
jtenner: These compile time and runtime checks should be in every function that accepts a `T` parameter.

packages/assembly/assembly/internal/RTrace.ts
jtenner: We can conditionally check to see if `T` is managed.

```ts
// short circuit for unmanaged
if (!isManaged<T>()) return 0;
```

packages/assembly/assembly/internal/RTrace.ts
9oelM: 👍

packages/assembly/assembly/internal/RTrace.ts
9oelM: You are right. Thanks :)

packages/core/README.md
9oelM: Thank you for catching that. 👍

packages/assembly/assembly/internal/RTrace.ts
9oelM: Thanks. I got that check separated into `private static` function.

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -76,6 +79,20 @@ export class RTrace {
    */
   public static enabled: bool = true;
 
+  /**
+   * This method checks if the type of the reference can be used.
+   * 
+   * @param {T} reference - the reference
+   * @param {string} getTarget - the information to get from the reference provided
+   */
+  private static assertReferenceType<T>(reference: T, getTarget: string): void {
```
jtenner: I know it says private, but the rest of my code style has static helper functions hoisted. Believe it or not, the standard library tries to avoid these because of wasm execution speed issues. However, in our case, we are trying to reduce compile times. I think refactoring that method into a private function is a great idea. However, we should hoist it out of the `RTrace` namespace.

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -220,23 +237,49 @@ export class RTrace {
...
+   * Get the type id (class id) of a reference.
+   *
+   * @param {T} reference - The reference.
+   * @returns {u32} - The type id of the allocated block.
+   */
+  public static typeIdOfReference<T>(reference: T): u32 {
+    if (!isManaged<T>()) return 0;
+    RTrace.assertReferenceType<T>(reference, "typeId");
+
+    return RTrace.typeIdOf(changetype<usize>(reference));
```
jtenner: We need to see what happens when this function is called. We may have to subtract `1` from the result. :)

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -220,23 +237,49 @@ export class RTrace {
...
-   * Get the size of a block or buffer.
+   * Get the size of a reference.
    *
    * @param {T} reference - The reference.
    * @returns {u32} - The size of the allocated block.
    */
-  public static sizeOf<T>(reference: T): u32 {
-    return load<u32>(changetype<usize>(reference) - 4);
+  public static sizeOfReference<T>(reference: T): u32 {
+    if (!isManaged<T>()) return 0;
```
jtenner: We need to do the checks here first before we return 0. Call `assertReferenceType()` first.

packages/assembly/assembly/internal/RTrace.ts
9oelM: Got it! :)

9oelM:
1. Tests started breaking because of the markdown length (I suspect they were breaking from some time ago already?)
```
README.md:558 MD013/line-length Line length [Expected: 80; Actual: 207]
README.md:568 MD033/no-inline-html Inline HTML [Element: T]
README.md:570 MD013/line-length Line length [Expected: 80; Actual: 96]
README.md:574 MD013/line-length Line length [Expected: 80; Actual: 191]
README.md:584 MD033/no-inline-html Inline HTML [Element: T]
README.md:586 MD013/line-length Line length [Expected: 80; Actual: 94]
README.md:590 MD013/line-length Line length [Expected: 80; Actual: 166]
README.md:592 MD033/no-inline-html Inline HTML [Element: T]
README.md:594 MD013/line-length Line length [Expected: 80; Actual: 107]
```
2. Need to test more (for throws, ...)

packages/assembly/assembly/internal/RTrace.ts
9oelM: yeap.

packages/assembly/assembly/internal/RTrace.ts
9oelM: Got it out of the class

packages/assembly/assembly/__tests__/RTrace.spec.ts
```diff
@@ -2,22 +2,80 @@ import { Vec3 } from "./setup/Vec3";
...
   it("should return a positive number", () => {
     globalVec = new Vec3(1, 2, 3);
     expect<i32>(RTrace.count()).toBeGreaterThan(0, "RTrace counts should be greater than 0");
   });
 
-  afterEach(() => {
-    let delta = RTrace.end(1);
-    expect<i32>(delta).toBe(1, "There should only now be one global reference to the Vec3 object");
+  it('should return a correct reference count of a pointer', () => {
+    let ptr: i32 = __alloc(offsetof<Vec3>(), idof<Vec3>());
```
jtenner: Remove the type annotation here. The type is actually `usize`. In fact remove as many type annotations as possible to reduce parse times.

```
expect<i32>(value);

// instead should be

expect(value);

Comment (9oelM) [reply packages/assembly/assembly/__tests__/RTrace.spec.ts]: Got it :) now I see why you are removing them from #184 as well.

**Comment by jtenner** [comment]
We can't really test compile time errors, not easily anyway. I'm going to open an issue and see if we can find a way to set up some jest tests and snapshots for that.

Just make sure it throws when `null` is passed to the reference functions.

```ts
throws("when null is passed to (function)", () =>{
  RTrace.functionName<Vec3 | null>(null);
}, "Null is not a valid parameter value for (function)");
```

packages/assembly/assembly/internal/RTrace.ts
jtenner: I'll get back to this too. Promise.

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -271,7 +297,32 @@ export class RTrace {
...
+  
+  /**
+   * Gets the current count of the specified reference.
+   * @param {T} reference - the reference.
+   */
+  public static refCountOfReference<T>(reference: T): u32 {
+    if (!isManaged<T>()) return 0;
+    assertReferenceType<T>(reference, "refCount");
+
+    return RTrace.refCountOf(changetype<usize>(reference));
```
jtenner: After careful inspection, it looks like any optimization level will remove any excess `__retain` function calls, and thus, the count returned by `refCountOfReference()` will be correct.

However, if the code is *not* optimized, then it returns `+1` which is technically correct, but also unexpected because we would want something like the following statement to return `1`.

```ts
let v = new TestClass();
// should always be 1 here by definition because it was just created
log(RTrace.refCountOfReference(b));
```

Thus the following snippet yields the correct function result. Note: The ternary operator will be inlined by the AssemblyScript compiler so there will be zero performance impact either.

```ts
  public static refCountOfReference<T>(reference: T): u32 {
    if (!isManaged<T>()) return 0;
    assertReferenceType<T>(reference, "refCount");

    let count = RTrace.refCountOf(changetype<usize>(reference));
    return ASC_OPTIMIZE_LEVEL > 0 ? count : count - 1;
  }
```

packages/assembly/assembly/internal/RTrace.ts
jtenner: This was not the correct function! Resolved.

packages/assembly/assembly/internal/RTrace.ts
jtenner: @9oelM Thank you for your effort. This particular fix is not something you should be expected to know when dealing with AssemblyScript, but I really appreciate all your help with these API changes. Please replace the source of the `refCountOfReference` function and I will inspect over this pull request one last time before pulling it.

packages/assembly/assembly/internal/RTrace.ts
```diff
@@ -271,7 +297,32 @@ export class RTrace {
...
+}
+
+/**
+ * iThis method checks if the type of the reference can be used.
+ * 
+ * @param {T} reference - the reference
+ * @param {string} getTarget - the information to get from the reference provided
+ */
+export function assertReferenceType<T>(reference: T, getTarget: string): void {
+  if (!isReference<T>()) ERROR("Cannot get " + getTarget + " when T is not a reference.");
```
jtenner: Ah. remove the `getTarget` parameter here, please. Don't worry about the property names because we cannot expect the compiler to concatenate strings dynamically for compile-time errors. Instead, manually inline the body of this function inside each calling function. This is unfortunate, but technically a correct way to do this.

Thank you for this change.

packages/assembly/assembly/internal/RTrace.ts
9oelM: @jtenner yay! thanks. I hope I will be able to know more for the next time I make changes!

packages/assembly/assembly/internal/RTrace.ts
9oelM: I remove the helper function and inlined them inside each of the functions we need to assert reference types. By the way, just out of curiosity, Is the reason I should not use `@inline`, that is going to have another side effect, as mentioned in #189?

PR #184, opened 2020-01-29
Support memory imports and generating function exports

### LINKED ISSUE #188: Show expected and actual value for string equality
When I compare two strings, the debugging output is not very helpful at the moment:

```ts
    it("strings are equal", () => {
      expect("a").toStrictEqual("b");
    });
```

prints

```
    ❌ strings are equal - 
      [Expected]: Reference Value
      [Actual]  : Reference Value
```

Would it be possible to create an output that helps understanding the problem (i.e. show expected and actual string)?

### LINKED ISSUE #189: Remove all instances of @inline to optimize module compile times

jtenner:
This is long overdue, and I'm glad I took my time to figure it out.

- [x] Add `--importMemory` flag to mandatory cli flags
- [x] support `imports` configuration as a function, and pass it the `WebAssembly.Memory` object
- [x] Add cli option for test module page size and page limit

Don't bother with docs. We need to figure out how to handle documentation better than the way we are currently doing it. Everything is disorganized and I'm not proud of the solution.

Time for an overhaul on the docs! 💪 

Fixes #189 
Fixes #188


packages/cli/src/util/CommandLineArg.ts
```diff
@@ -189,6 +190,13 @@ const _Args: CommandLineArgs = {
     value: false,
   },
 
+  memory_size: {
+    description: "Set the initial size of imported memory in bytes.",
```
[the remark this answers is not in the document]
jtenner: I just changed this to `--memory-max`  (default is -1) and `--memory-size` which is now in web assembly pages.

jtenner:
Waiting for review on @willemneal

packages/cli/src/help.ts
```diff
@@ -18,6 +18,8 @@ export function help() {
     {bold.green asp} -h
     {bold.green asp} --types                         Copy the types file to assembly/__tests__/as-pect.d.ts
     {bold.green asp} -t
+    {bold.green asp} --memory_size=[number]          Initial size of imported memory in bytes. {yellow (Default: 10 Bytes)}
```
willemneal: This was in 64 kB pages right?

packages/cli/src/util/CommandLineArg.ts
```diff
@@ -189,6 +191,19 @@ const _Args: CommandLineArgs = {
     value: false,
   },
 
+  "memory-max": {
```
willemneal: This is missing from the help doc.  Perhaps a future PR can craft the help doc from the definitions.

packages/cli/src/run.ts
```diff
@@ -390,11 +394,17 @@ export function run(cliOptions: Options, compilerArgs: string[]): void {
...
+
+      const configurationImports = fs.existsSync(customImportFileLocation)
+        ? require(customImportFileLocation)
+        : configuration!.imports ?? {};
+
+      const memory = new WebAssembly.Memory({ initial: 10 });
+      const stagedImports = typeof configurationImports === "function"
+          ? configurationImports(memory)
+          : configurationImports;
+      const imports = runner.createImports(stagedImports);
```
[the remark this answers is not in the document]
willemneal: I think that we should change it, but perhaps it's okay for now.

packages/cli/src/util/CommandLineArg.ts
jtenner: Yeah that sounds right. I'm currently overhauling the primary docs to use gitbook. I'll fix the cli options text today though.

PR #192, opened 2020-02-08
[Feature] implement proper toStrictEqual

jtenner:
Goal: Implement `refa.__aspectStrictEquals(refb, stackA, stackB)` function on each class prototype. 

CC @willemneal and @9oelM


jtenner:
This all passes now with a simple nested reference 

Notes:

1. This is a proof of concept and the AST source transformer will need to be refactored to make it easier to understand.
2. We can utilize this technique to help make logging references WAY more user-friendly for version `3.0`
3. Need more testing examples and we should worry about object logging and reporting for later, because it will definitely break the way references are currently reported
  - reporting values when expectations aren't met means querying into wasm and we lose type info
  - need to store the object outside of wasm *first* before it can be referenced later. (Very slow, need to investigate)
  - logging values should be a piece of cake

jtenner:
cc @MaxGraey

jtenner:
Things to do:

- [ ] Implement proper object logging using a object builder state machine (more calls into js yay)
- [ ] Fix docs
- [ ] Add more tests
    - [ ] Test circular references
    - [ ] Test deeper references
    - [ ] Test negatives/contrapositives/negative contrapositives
    - [ ] Test nullable reference fields with negatives/contrapositives/negative contrapositives

jtenner:
As it turns out, class members can be either `Public` `Private` or `Protected`. But they can also be marked as none of these, in whcih case, public is assumed by default. This actually makes the proof of concept work, mostly.

jtenner:
This issue should be fixed now. Gonna use the nightly compiler and see if that helps me get `Set` and `Map` working.

jtenner:
Travis says this is failing, but that's because of this silly v8 null pointer exception that can't be fixed easily apparently. 

Ignore travis. The tests pass.

packages/assembly/assembly/__tests__/toInclude.spec.ts
```diff
@@ -113,4 +119,32 @@ describe("toInclude reference arrays", () => {
   throws("should include a function pointer", () => {
     expect(eventDispatcher.events).not.toInclude(listener);
   });
+
+  /**
+   * Sets should be value actual targets for toInclude function calls.
```
willemneal: actual values?

packages/assembly/assembly/__tests__/toIncludeEqual.spec.ts
```diff
@@ -151,3 +156,33 @@ describe("arrays with strings", () => {
     expect(stringRefs).toIncludeEqual("four");
   }, "The value is not included.");
 });
+
+describe("sets", () => {
+  /**
+   * Sets should be value actual targets for toInclude function calls.
```
willemneal: see above

willemneal:
Also looks like it's the weird node issue again...

jtenner:
One step closer to `3.0.0`

Issue #203, opened 2020-02-14
Remove generic parameters to instanceof checks


jtenner:
https://github.com/jtenner/as-pect/blob/master/packages/assembly/assembly/internal/comparison/toIncludeComparison.ts#L42

Also in toIncludeEqualsComparison 

Anywhere else in the code where this might occur

PR #208, opened 2020-02-16
Replace generic parameters with instanceof checks (#203)

9oelM:
- [x] Replace generic parameters with instanceof checks, according to https://github.com/AssemblyScript/assemblyscript/pull/1108
[ ] ~Add TypedArray check~: Leaving it for later

Fixes: #203


packages/assembly/assembly/internal/comparison/toIncludeComparison.ts
```diff
@@ -39,7 +39,7 @@ export function toIncludeComparison<T, U>(actual: T, expected: U, negated: i32,
   let includes = false;
 
   // test for Sets
-  if (actual instanceof Set<indexof<T>>) {
+  if (actual instanceof Set) {
     // @ts-ignore: type safe .has(expected) method call
```
jtenner: We can remove this ts-ignore line now.

packages/assembly/assembly/internal/comparison/toIncludeEqualComparison.ts
```diff
@@ -40,7 +40,7 @@ export function toIncludeEqualComparison<T, U>(actual: T, expected: U, negated:
   let includes = false;
 
   // test for Sets
-  if (actual instanceof Set<indexof<T>>) {
+  if (actual instanceof Set) {
     // @ts-ignore: type safe .has(expected) method call
```
jtenner: Remove this ts-ignore too

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -60,7 +60,7 @@ export class Reflect {
 
       if (isDefined(left[0])) { // test for safe indexof usage
```
jtenner: We can remove this `isDefined` check now because we can garuntee Set and Map are in the prototype chain.

jtenner:
Thank you so much!

PR #214, opened 2020-02-19
Allow multiple cli arguments to be passed and fix null field values

willemneal:
This is needed if a project wants to add its own transformer.  This adds a test of an empty transformer. Also this allows for arguments to be strings or arrays.


packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -49,9 +49,10 @@ export class Reflect {
...
-    // if the value is null, create a Null host value
-    if (isNullable<T>()) {
-      if (value == null) {
+    // if T is a reference
+    if (isReference<T>()) {
+      // if the value is null, create a Null host value
+      if (changetype<usize>(value) == 0) {
         return createHostValue(
           true,
           false,
```
jtenner: we don't know if the reference type is `nullable` can we add a call to `isNullable<T>()` in the `nullable` argument?

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -316,7 +313,7 @@ export class Reflect {
     return 0;
   }
 
-  public static equals<T>(left: T, right: T, stack: usize[] = [], cache: usize[] = []): i32 {
+  public static equals<T>(left: T, right: T, stack: usize[] = [], cache: usize[] = []): i32 {
```
jtenner: leading whitespace nit

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -333,7 +330,7 @@ export class Reflect {
     // if it's possible for T to be null
     if (isNullable<T>()) {
       // mutual exclusion null
-      if (i32(left == null) ^ i32(right == null)) return Reflect.FAILED_MATCH;
+      if (i32(changetype<usize>(left) == 0) ^ i32(changetype<usize>(right) == null)) return Reflect.FAILED_MATCH;
```
jtenner: check for `== 0` here

packages/cli/src/run.ts
```diff
@@ -379,7 +379,7 @@ export function run(cliOptions: Options, compilerArgs: string[]): void {
 
   // create the array of compiler flags from the flags object
   const flagList: string[] = Object.entries(flags)
-    .reduce((args: string[], [flag, options]) => args.concat(flag, options), [])
+    .reduce((args: string[], [flag, options]) => args.concat(flag, options.length == 0 || (typeof options == "string") ? options : options.join(",")), [])
```
jtenner: Seriously. Thank you for this fix. I'm glad it helps.

packages/core/src/transform/assemblyscript.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from "assemblyscript";
+export * from "assemblyscript/cli/transform"
```
jtenner: nit: add leading newline here :)

packages/core/src/transform/emptyTransformer.ts
```diff
@@ -0,0 +1,12 @@
+import {Transform, Parser} from "./assemblyscript";
```
jtenner: Nit: `import { Transform, Parser }` whitespace

packages/core/src/transform/emptyTransformer.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export = class AspectTransform extends Transform {
+  /**
```
jtenner: We can remove this comment and simply comment the class that it's an example transform for testing purposes.

packages/assembly/assembly/internal/Reflect.ts
jtenner: Reason: Compiler warning

PR #215, opened 2020-02-19
add isManaged, handle unmanaged classes

jtenner:


jtenner:
Ignore pikaci. Travis should pass.

PR #222, opened 2020-02-20
[Refactor] Deprecate expectFn and put warning in toThrow

### LINKED ISSUE #213: Deprecate expectFn
Because type inference is now a thing!

9oelM:
- [x] @deprecate expectFn
- [x] put warning in `toThrow`
- [x] just change examples in README.md

would close #213


packages/assembly/assembly/__tests__/unreachable.spec.ts
```diff
@@ -35,7 +35,7 @@ describe("unreachable", () => {
    * This test validates that a negated toThrow assertion is valid.
    */
   it("should run normally", () => {
-    expectFn(() => {
+    expect(() => {
       counter = 1;
     }).not.toThrow("unreachable instructions should throw.");
   });
```
9oelM: did not change the tests for `expectFn` below because they are deprecated but still supported. Is it a good approach?

jtenner: We should not use our own deprecated apis. Please update those tests to use expect as normally.

9oelM: updated them :)

PR #223, opened 2020-02-20
Add DisplayAs method to override host value reporting

DuncanUszkay1:
We wanted to be able to override the what was returned by toHostValue calls in the Reflect module. This is useful when you have an class which has an interpretation that doesn't exactly match the properties of that class.


jtenner:
I'm going to merge a request involving snapshots of logging `HostValue` output. So you'll have to rebase and update the snapshots. If you run into trouble let me know.

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -77,7 +77,13 @@ export class Reflect {
           return seen.get(changetype<usize>(value));
         }
       }
-      if (value instanceof ArrayBuffer) {
+      if (isDefined(value.__aspectDisplayAs())) {
+        let display_value = value.__aspectDisplayAs();
+        if (!isInteger(display_value) && !isFloat(display_value) && !isManaged(display_value)) {
```
willemneal: Since we have the type variable, it's better practice to add it to each generic function.  I know it's probably an old fear of being burned my type inference in earlier versions of  the compiler, but I also need to check to see if adding the generic arguments generates better code since it should be a compile time constant.  It technically should be the same, but I'm just wondering.

DuncanUszkay1: I don't think we do- we have T, but here we're checking the type of T.__aspectDisplayAs() which does not have to be, and almost certainly won't be T.

jtenner: We don't have the type variable which would be the return value of a given `__aspectDisplayAs()` function. However, we can garuntee that it exists and returns anything other than `void` if `isDefined()` works properly. The `Reflect.toHostValue<T>()` function will easily infer what `__aspectDisplayAs()` returns and we can safely treat it as `any`, even though the compile knows what that value will be.

This technique absolutely relies on type inference, and it very elegant.

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -77,7 +77,13 @@ export class Reflect {
           return seen.get(changetype<usize>(value));
         }
       }
-      if (value instanceof ArrayBuffer) {
+      if (isDefined(value.__aspectDisplayAs())) {
+        let display_value = value.__aspectDisplayAs();
```
jtenner: Would prefer camelCase variable names if I can help it.

`displayValue`

PR #233, opened 2020-02-21
Support Snapshots📸

jtenner:
The idea here is that we can use the `diff` package to aid in creation of a `Snapshot` and `SnapshotComparison` objects that can `diff()` and generate a set of `SnapshotDiff` objects.

The aspect cli can then determine if it needs to `exit 1` or `0` properly given diff output.

TODO:

- [x] create snapshot package
- [x] snapshot the snapshot output (The IRONY)
- [ ] add `TestCollector` support for collecting snapshots using the new `Reflect` api in the `assembly` package
- [ ] add cli check to load/save snapfiles to `${dirname}/artifacts/${basename}.ts.snap`
- [ ] add cli check to save watfiles to `${dirname}/artifacts/${basename}.wat`
- [ ] add `TestContext` support for diffing snapshots if they were collected or provided
- [ ] `process.exit(1)` if snapshots differ
- [ ] add snapshots that were added programatically and save them automatically
- [ ] add support for viewing diffs in `VerboseReporter`
- [ ] add support for viewing stats in `SummaryReporter`
- [x] avoid calls to `__aspectDisplayAs()` function


jtenner:
100% code coverage with current snapshots, ironically. It's probably fine to leave the tests like this because the module is small and generic.

jtenner:
Next step is to add a `toMatchSnapshot(name)` method. It should call out to `reportExpectedSnapshot(actualHostValue, snapshotName, negated)`

jtenner:
There is not a good way to complete this pull request with my sanity intact before I refactor the way test collection works.

PR #237, opened 2020-02-22
Bug fixes and cleanups

### LINKED ISSUE #227: Better express negated test failures
When a negated expectation fails like:
```
let a = new Array<i32>();
a = [1,2,3]
expect<Array<i32>>(a).not.toStrictEqual(a)
Right now we output:
❌ assembly/__tests__/example.spec.ts Pass: 0 / 1 Todo: 0 Time: 2.858ms
  Failed: example
    ❌ should display the override method -
      [Expected]: Array<i32> [1, 2, 3, ]
      [Actual]  : Array<i32> [1, 2, 3, ]
```
But probably we should output:
```
❌ assembly/__tests__/example.spec.ts Pass: 0 / 1 Todo: 0 Time: 2.858ms
  Failed: example
    ❌ should display the override method -
      [Not Expected]: Array<i32> [1, 2, 3, ]
      [Actual]  : Array<i32> [1, 2, 3, ]
```
Or something that indicates that the test is a negation.

### LINKED ISSUE #228: inline arrays have leading commas
When we have the following code:
```
expect<Array<i32>>([1,2,3]).toBe([3,4,5])
```
We get this output:
```
❌ assembly/__tests__/example.spec.ts Pass: 0 / 1 Todo: 0 Time: 4.388ms
  Failed: example
    ❌ should display the override method -
      [Expected]: Array<i32> [3, 4, 5, ]
      [Actual]  : Array<i32> [1, 2, 3, ]
```
This output has an extra comma which ought to be removed. The intended output looks like this:
```
❌ assembly/__tests__/example.spec.ts Pass: 0 / 1 Todo: 0 Time: 4.388ms
  Failed: example
    ❌ should display the override method -
      [Expected]: Array<i32> [3, 4, 5]
      [Actual]  : Array<i32> [1, 2, 3]
```

### LINKED ISSUE #231: Rework abort error message
Found this while writting a fairly minimal example:
```typescript
class Parent {
  constructor(public prop: u8) {
  @inline
  static fromProp(prop: u8): Parent { return new Parent(prop) }
}

class Child extends Parent {
  @inline
  static fromProp(prop: u8): Child {
    return <Child>Parent.fromProp(prop)
  }
}
describe("example", () => {
  it("should match when using non-constructor", () => {
    expect<Child>(Child.fromProp(1)).toStrictEqual(Child.fromProp(2));
  });
});
```
It yields this output:
```typescript
❌ assembly/__tests__/example.spec.ts Pass: 0 / 1 Todo: 0 Time: 3.169ms
  Failed: example
    ❌ should match when using non-constructor - No assertion message provided.
```
Looks like it came from an abort that I didn't explicitly call:
```typescript
/**
 * This function overrides the provided AssemblyScript `env.abort()` function to catch abort
 * reasons.
 *
 * @param {number} reasonPointer - This points to the message value that causes the expectation to
 * fail.
 * @param {number} _fileNamePointer - The file name that reported the error. (Ignored)
 * @param {number} _line - The line that reported the error. (Ignored)
 * @param {number} _col - The column that reported the error. (Ignored)
 */
TestCollector.prototype.abort = function (reasonPointer, _fileNamePointer, _line, _col) {
    this.message = this.getString(reasonPointer, "No assertion message provided.");
};
```
Unfortunately the message provided only makes sense in the context of an abort that results from an assertion, like this:
```typescript
describe("example", () => {
  it("should match when using non-constructor", () => {
    assert(false)
  });
});
```
We should change this message such that we get all the necessary debug information for fixing the issue that caused the abort, without making the error message for assertions with no messages look confusing

### LINKED ISSUE #234: Add lib folders to .gitignore
Since the lib folders are published with the npm package, there is no need to also include the files in the git repo.

### LINKED ISSUE #238: Deprecate Performance API
We can start development on a performance measurement cli with it's own core and assembly packages. Since compiling a module with optimizations is slow, it doesn't make sense to mingle test collection/execution with performance testing.

As a result, I will push a `Warning` for every test that has `Performance.enabled(true)` when the test runs so that these warnings get reported to the console.

jtenner:
Fixes #238 
Fixes #234 
Fixes #231 
Fixes #227 
Fixes #228


jtenner:
At @DuncanUszkay1 request, I'll move prettier to root and run it once on the entire code base with a prettier ignore for the single file in the assembly package.

PR #241, opened 2020-02-23
Snapshot reporters output

### LINKED ISSUE #240: Snapshot testing for VerboseReporter  and SummaryReporter
- [ ] Create a generic module to test stdout  (in jest test, we can bootstrap our own stdout and snapshot the output)
- [ ] Make sure chalk is disabled and doesn't output ansii, or strip ansii from result
- [ ] Loop over each test and log item, snapshot each individual group and test

jtenner:
Preliminary exploratory setup for snapshotting output appropriately.

Fixes #240


jtenner:
Some branches simply check for value truthiness and are sanity checks to make sure output looks visually stable.

Now I need to snapshot Summary reporter

jtenner:
This is completed to the best of my ability. For now, if reporter output changes for any reason, it will be captured in the snapshots and can be inspected, visually, by the developer. This is a great use of snapshots in general.

packages/core/src/reporter/SummaryReporter.ts
```diff
@@ -15,18 +15,30 @@ export default class SummaryReporter extends TestReporter {
...
+      if (
+        !options.enableLogging ||
+        /* istanbul ignore next */ options.enableLogging === "false"
+      )
+        /* istanbul ignore next */
         this.enableLogging = false;
     }
   }
 
+  /* istanbul ignore next */
```
DuncanUszkay1: What's up with these

jtenner: `options` can be null, and it's only to set the `enableLogging` property. I opted to just ignore testing for this, because it's just a bunch of sanity checks, and I am only testing code coverage of parts of this software actually reponsible for logging output.

jtenner: Also, istanbul is annoying.

packages/core/__tests__/__snapshots__/SummaryReporter.spec.ts.snap
```diff
@@ -0,0 +1,61 @@
...
+"❌ assembly/jest-reporter.ts Pass: 4 / 7 Todo: 1 Time: 0ms
+  Failed: examples
+     [Log]: \\"testing in a group\\"
+    ❌ something fails - Here is an example failure message.
+      [Expected]:  42
+      [Actual]  : -1
+    ❌ this should throw but it does not - it should throw
+    ❌ something negated fails - 42 is 42
+      [Expected]: Not  42
+      [Actual]  : 42
```
9oelM:
```
      [Expected]: Not  42
```
This part has two spacings between Not and 42, while I think there should be one? maybe?

packages/assembly/README.md
```diff
@@ -37,7 +37,8 @@ following AssemblyScript `asc` cli options:
 ```
 asc entry.ts ./node_modules/@as-pect/assembly/assembly/index.ts \
   --use ASC_RTRACE=1 \
-  --explicitStart
+  --explicitStart \
+  --transform ./node_modules/@as-pect/core/lib/transform/index.ts
```
willemneal: This shouldn't this be the `--transform @as-pect/core/lib/transform`?

jtenner: Hey does that work? I wasn't aware!

jtenner: These docs will be available on gitbook, so I'm not worried about this. Ignoring for now :)

packages/core/__tests__/__snapshots__/SummaryReporter.spec.ts.snap
jtenner: Oh whoops. Yeah that seems like something to fix. I'll handle it later.

PR #242, opened 2020-02-24
Fix toBeNull assertions

jtenner:
It's possible for nulls to exist, masquerading as references that are uninitialized. This removes the compile time check for `isNullable()` and compile time errors, in favor of `actual instanceof usize` and `isReference(actual)`.  Update `as-pect` test suite, too.


packages/assembly/README.md
```diff
@@ -37,7 +37,8 @@ following AssemblyScript `asc` cli options:
 ```
 asc entry.ts ./node_modules/@as-pect/assembly/assembly/index.ts \
   --use ASC_RTRACE=1 \
-  --explicitStart
+  --explicitStart \
+  --transform ./node_modules/@as-pect/core/lib/transform/index.ts
```
willemneal: `@as-pect/core/transform`, but might require `/lib/transform` instead

PR #244, opened 2020-02-25
Support instantiation of module from config

### LINKED ISSUE #232: Expose the current instance to custom imports
Currently custom imports can be created with a function that takes a reference to the current instance's memory, but it would be even more useful if it got access to the fully constructed ASLoader constructed instance.  This would allow for using all of the exports and use the convenience functions that allow for constructing arbitrary objects and arrays, which is super useful for imports that want to inject data into the tests.

This could be solved in several ways.  

1. In TextContext before running the instance, set it to a global (not preferred, but works currently with testing)
2. Have TextContext.run be responsible for instatiating the module and then passing a reference to itself when constructing custom imports.
3. Have the custom imports be an exported class and constructing the imports would require returning an object where each import methods are bound to the class.  Then after the module is created pass the class instance a reference to it.

Personally I'm down for the third option, which could still be done with old fashioned non-class js.

jtenner:
fixes #232 

This remains backwards compatible for everyone except for those who started to use the `WebAssembly.Memory` with the imports callback. Now those functions require that the configuration function returns an AssemblyScript module instead of the imports.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,10 +11,10 @@
       "version": "file:packages/cli",
       "dev": true,
       "requires": {
-        "@as-pect/assembly": "^3.0.0-alpha.3",
-        "@as-pect/core": "^3.0.0-alpha.3",
-        "@as-pect/csv-reporter": "^3.0.0-alpha.3",
-        "@as-pect/json-reporter": "^3.0.0-alpha.3",
+        "@as-pect/assembly": "^3.0.0-beta.0",
+        "@as-pect/core": "^3.0.0-beta.0",
+        "@as-pect/csv-reporter": "^3.0.0-beta.0",
+        "@as-pect/json-reporter": "^3.0.0-beta.0",
         "chalk": "^3.0.0",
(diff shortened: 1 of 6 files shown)


willemneal:
Looks good, but perhaps it would be better to pass the following
 ```ts
let createImports = runner.createImports.bind(runner);
```

instead of the imports so that the user can use it and not have to deal with the merging of the imports.

**Comment by jtenner** [comment]
I like this idea. Not perfectly sure that's the right solution, but it's less mental work for the user.

**Comment by jtenner** [comment]
```js
module.exports = function (_memory, createImports, instantiateSync, binary) {
  return instantiateSync(binary, createImports({
    api: {
      identity(input) { return input; },
    }
  }));
};
```

jtenner:
@DuncanUszkay1  and @willemneal 

https://app.gitbook.com/@tenner-joshua/s/as-pect/cli-configuration

Can you help me with the docs here? I want to make sure this is understandable.

PR #247, opened 2020-02-28
Refactor TestResult and TestGroup into TestNode

jtenner:
This is turning out to be a literal backward-incompatible rewrite involving removing the performance API, refactoring the way the RTrace API works, and pulling lots of my hair out.

- [X] Remove Performance API
- [X] Refactor RTrace API
- [X] Fix tests
- [X] New snapshots
- [ ] Test output of deeply nested tests
- [x] Refactor CLI
- [x] Do output cleanup
- [X] Bug fixes
- [x] Remove Expectation.ready logic (top level expectations should work)


willemneal:
This won't affect the user level API's for tests though correct? (if they weren't using performance APIs)

jtenner:
That is correct. Test level stuff should be backwards compatible. Anyone using reporters (nobody afact was using them) will see compatibility issues.

packages/core/src/test/TestContext.ts
```diff
@@ -1,437 +1,1415 @@
...
+      this.groupCount += 1;
+      if (pass) this.groupPassCount += 1;
     } else {
-      this.runBeforeEach(group, result);
-      if (this.endGroup) return;
-      this.runTestCall(group, result);
-      this.runAfterEach(group, result);
-      if (this.endGroup) return;
+      this.testCount += 1;
+      if (pass) this.testPassCount += 1;
```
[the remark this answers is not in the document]
jtenner: The properties are `this.testCount` and `this.testPassCount` for test statistics.

jtenner:
Currently there are still visual problems with the output, and some lines which aren't covered, and I'm unsure what tests need to be written.

Things to note:

1. There are no warnings that aspect reports any more. Any lines that report warnings can't be covered. (hence lots of line ignoring) 
2. I need a break
3. In its current state, everything seems to be backwards compatible for everyone except for those who had custom reporters and used the performance api.
4. Expectations can now happen inside describe blocks (and at the top level outside of a test or a describe) 
5. Control flow works (beforeEach, etc.) 
6. The method that collects the tests also RUNS the test as well, so the TestCollector class was rendered useless
7. Invalid expectation calls are no longer a thing, and the jest tests for those issues were removed as a result
8. Jest compatibility was increased by quite a lot. 

At this point we have two options. 

1. We can push the unfinished software under an alpha build (and get it in the hands of our users quickly to help get errors reported faster) 
2. We can keep striving for 100% line coverage and wait longer to push

jtenner:
Ping @willemneal @MaxGraey @DuncanUszkay1

README.md
```diff
@@ -1,7 +1,6 @@
 # jtenner/as-pect
 
-This packages is a monorepo that contains the cli and the core for the `@as-pect`
-packages.
+This packages is a monorepo that contains the cli and the core for the `@as-pect` packages.
```
willemneal: ~This packages~ This repository, or This package

packages/cli/src/util/collectReporter.ts
```diff
@@ -13,11 +9,11 @@ import chalk from "chalk";
...
  * @param {Options} cliOptions - The command line arguments.
  */
-export function collectReporter(cliOptions: Options): TestReporter {
-  const reporters: TestReporter[] = [];
+export function collectReporter(cliOptions: Options): IReporter {
+  const reporters: IReporter[] = [];
 
   if (cliOptions.csv) {
-    const CSVReporter = require("@as-pect/csv-reporter").default;
+    const CSVReporter = require("@as-pect/csv-reporter").CSVReporter;
```
willemneal: Can't this be `exports = CSVReporter`, in its file and then not have to reference it here?

packages/cli/src/util/collectReporter.ts
```diff
@@ -30,7 +26,7 @@ export function collectReporter(cliOptions: Options): TestReporter {
   }
 
   if (cliOptions.json) {
-    const JSONReporter = require("@as-pect/json-reporter").default;
+    const JSONReporter = require("@as-pect/json-reporter").JSONReporter;
```
willemneal: Same as above.

packages/core/__tests__/TestGroupFilter.spec.ts
```diff
@@ -24,24 +25,33 @@ describe("TestGroup filtering", () => {
...
     // @ts-ignore setting the protected testRegex property is just for testing
     context.testRegex = /two/i;
     context.run(wasm);
-    expect(context.testGroups).toHaveLength(3);
 
-    context.testGroups.forEach(group => {
-      expect(group.tests).toHaveLength(1);
+    let count = 0;
+    context.rootNode.visit(node => {
+      if (node.type === TestNodeType.Test && node.ran) count += 1;
```
willemneal: Copied code. Perhaps this could be a method on the node.

packages/core/src/reporter/SummaryReporter.ts
```diff
@@ -1,20 +1,17 @@
...
 import { IWritable } from "../util/IWriteable";
 import { ReflectedValue } from "../util/ReflectedValue";
+import { IReporter } from "./IReporter";
 
 /**
  * This test reporter should be used when logging output and test validation only needs happen on
  * the group level. It is useful for CI builds and also reduces IO output to speed up the testing
  * process.
  */
-export default class SummaryReporter extends TestReporter {
```
willemneal: So what is the new way to pass custom reporters?  Before we could assume that it would be the default export.

packages/core/src/reporter/SummaryReporter.ts
```diff
@@ -41,7 +44,8 @@ export default class SummaryReporter extends TestReporter {
   /* istanbul ignore next */
   public onTodo(): void {}
 
-  private stdout: IWritable | null = null;
+  public stdout: IWritable | null = null;
```
willemneal: Shouldn't summary and verbose reporters be sub-classes of an abstract one that then handles the I/O setup, or anything that they both share?

packages/core/__tests__/TestGroupFilter.spec.ts
jtenner: Another thought might be to add a `ranCount` property to `TestContext`. Remember, I'm still trying to figure out what is useful.

packages/core/src/reporter/SummaryReporter.ts
jtenner: Oh. You're right. We can still use the `default` export for custom ones, but for ours, we can use this way. I just like the code style better.

packages/core/src/reporter/SummaryReporter.ts
jtenner: If we want to compile them to Web Assembly at some point, we will have to.

packages/core/__tests__/TestGroupFilter.spec.ts
jtenner:
```ts
  /** The test count. */
  public testCount: number = 0;

  /** The number of tests that ran. */
  public testRunCount: number = 0;

  /** The number of passing tests count. */
  public testPassCount: number = 0;

  /** The group count. */
  public groupCount: number = 0;

  /** The number of groups that ran. */
  public groupRunCount: number = 0;

  /** The number of passing groups count. */
  public groupPassCount: number = 0;
```

packages/core/src/reporter/SummaryReporter.ts
jtenner: The `EmptyReporter` has a bunch of empty method stubs for this purpose, and I think that forcing our users to create and extend a class when a simple object will do is not necessary. We can revisit this later for `4.0`.

packages/core/__tests__/TestGroupFilter.spec.ts
willemneal: Another idea is to collect the nodes as they run.

jtenner: I'll explain on slack.

Issue #149, opened 2019-09-01
[Add] DOM reporter (maybe even react support) in browsers


jtenner:

9oelM:
What's up with this one?

jtenner:
It's useful to run tests in a browser, and thus should be able to report the test results to a web page. I don't have any use cases specifically, or people who want this feature because setting up browser tests is a manual thing, but it's something I think about.

9oelM:
I think we need to discuss more on this

PR #253, opened 2020-03-03
Implement DomReporter

9oelM:
~Before v.3.1.0-alpha, I had a problem with json-reporter and csv-reporter which complained about modules that can only be used for nodejs environment (non-browser).
That error's gone because now those reporters are optional, so that's good~

~But I think there's still a problem with the export somehow, because parcel/storybook/CRA complain all about the import (I tried all of these three options just in case something is wrong with the tool), like below:~

image

~One more problem was perf_hooks being used in `run.ts`, but I think parcel auto-resolves this module to the one that supports browser. So this one doesn't matter.~ <- ~This is only supported in parcel 2 and parcel 2 is still in its alpha.~ 
 
Hopefully we will find a solution as we investigate it.

Todos:
- [x] ~Use browser's native `window.performance` instead of `perf_hooks` when it is run on browser~ Use isomorphic module instead
- [x] ~Fix `baseUrl` in `tsconfig` to be able to read from `lib/index.js` to make the import on browser possible~
- [x] ~tsconfig for cli/browser: should separate?~ resolved in #258 
- [ ] Implement DomReporter (tasks written in #149)


jtenner:
The whole goal of this is to avoid using the cli.

We should be able to only use the core package in the browser.

9oelM:
@jtenner got it :)

jtenner:
I think that the files are transpiled down to es5 node.js modules. Is there a way to transpile them down to proper js modules so parcel can use them? I don't know if that breaks node 12 compatibility.

9oelM:
I got it working, and I will push it after I'm done with my work for today. :) 
@jtenner

jtenner:
@9oelM I think we will only need `commonjs` and I'll have that change lading on master soon.

9oelM:
image
Got browser imports working, but still several points to work out:
- [ ] find out why `WebAssembly.instantiate` is not working but `instantiate` from assemblyscript loader does
- [ ] find a better way to compile and deal with `NameSection.buffer`, because 
```js
  constructor(contents: Uint8Array) {
    const mod = new WebAssembly.Module(contents);
    const section = WebAssembly.Module.customSections(mod, "name")[0];
    this.section = new WasmBuffer(new Uint8Array(section));
    this.parseSection();
  }
```
had to be changed to
```js
    WebAssembly.compile(contents).then((mod) => {
      const section = WebAssembly.Module.customSections(mod, "name")[0];
      this.section = new WasmBuffer(new Uint8Array(section));
      this.parseSection();
    });
```
because the browser would give us an error if we directly initialize a wasm module of a size more than 4kb with `WebAssembly.Module` because it doesn't want to block the main thread. But putting promise in the constructor might not be a good idea.
- [ ] Do `WhateverNameReporter implements EmptyReporter`
- [ ] Make React component to consume output from `WhateverNameReporter`
- [ ] Get it working with web worker

jtenner:
- [ ] create a single component that passes the binary to a single web worker (dedicated for running binaries)

If people want to run multiple binaries, they can instantiate more components. We can use setTimeout task scheduling. 

- [ ] add TestContext async option (default: false) to avoid breaking synchronous cli instantiation

In the case that we need to instantiate async in the browser, this option need not be passed to our end users. We are going to use promises internally. 

- [ ] visitAsync algorithm (I need to work on this) 

We need a way to cancel the execution of a test binary from the browser side, so between each node execution we should check to see if TestContext execution was "canceled."

Perhaps this will require a change to the Core api.

jtenner:
Hey @9oelM I hope everything is well. Should I just close this pull request?

9oelM:
Hey, i'm sorry that it's been a long overdue work.. Yes, let's leave it closed for now.

PR #254, opened 2020-03-04
Implement Calls to super.__aspectStrictEquals and comments, cleanups

### LINKED ISSUE #250: toStrictEquals on extended classes
toStrictEquals uses the Reflect module to provide object traversal equality. However, two objects differ only in fields that exist on their parent classes, this comparison will return true.

For Example:
```typescript
@unmanaged class Parent {
  id: u32
  constructor(x: i32) {
    this.id = x
  }
}
@unmanaged class Child extends Parent {
  static from(x: i32): Child { return <Child>(new Parent(x)) }
}
describe("Child", () => {
    it("children with different ids dont strict equal", () => {
        expect<Child>(Child.from(2)).toStrictEqual(Child.from(1));
    });
});
```
This test should not pass, but does.

jtenner:
- [x] check if each property needs to be ignored
- [x] check if `isDefined(super.__aspectStrictEquals())`
- [x] implement call to `super.__aspectStrictEquals()`
- [x] Apply same fixes for keyValuePairs in object classes
- [x] Update snapshots
- [x] fix method declaration inspection

Fixes #250


packages/core/src/transform/createStrictEqualsMember.ts
```diff
@@ -163,34 +174,60 @@ function createStrictEqualsFunctionBody(
...
+      ),
+      null,
+      // ("prop")
+      [TypeNode.createStringLiteralExpression(name, range)],
+      range,
+    ),
+    range,
+  );
+
   // if (Reflect.equals(this.prop, ref.prop, stack, cache) === Reflect.FAILED_MATCH) return false;
```
9oelM: would be good if the comments here are changed too

jtenner: Changed to what?

9oelM: Oh. I thought it has to be `if (Reflect.equals(this.prop, ref.prop, stack, cache) == Reflect.FAILED_MATCH) return false;` as well (equalsequals).

jtenner: The __aspectStrictEquals function returns a boolean because there are only two possible outcomes. When comparing a property, there are three outomes because it's possible for a circular reference to occur.

packages/core/src/transform/createAddReflectedValueKeyValuePairsMember.ts
```diff
@@ -104,7 +105,7 @@ function createAddReflectedValueKeyValuePairsFunctionBody(
 ): BlockStatement {
   const body = new Array<Statement>();
   const range = classDeclaration.name.range;
-  const names = new Array<string>();
+  const names = new Array<number>();
```
DuncanUszkay1: Minor nitpick- I think this name and the argument `ignore` should indicate that they are hashes. Something like `nameHashes` and `ignoreHashes` maybe. Just makes it clear that the names are not in fact being stored, but just their hash

packages/core/src/transform/hash.ts
```diff
@@ -0,0 +1,7 @@
+export function hash(str: string): number {
+  const points = Array.from(str);
+  let h = 0;
+  for (let p = 0; p < points.length; p++)
+    h = 37 * h + points[p].codePointAt(0)!;
+  return h;
+}
```
DuncanUszkay1: Is there no existing hashing package we can use instead of rolling our own?

packages/core/src/transform/createAddReflectedValueKeyValuePairsMember.ts
```diff
@@ -207,7 +212,7 @@ function createIsDefinedIfStatement(names: string[], range: Range): Statement {
                   // [...propNames]
                   TypeNode.createArrayLiteralExpression(
                     names.map(e =>
-                      TypeNode.createStringLiteralExpression(e, range),
+                      TypeNode.createIntegerLiteralExpression(f64_as_i64(e), range),
```
DuncanUszkay1: e feels like an odd choice of variable name to represent a hash.

Also our hash spitting out a f64 is very odd

packages/core/src/transform/hash.ts
jtenner: I like how your criticism was laid. Do you have any suggestions? I'm very inexperienced in these sorts of matters.

packages/core/src/transform/createAddReflectedValueKeyValuePairsMember.ts
jtenner: Which function should I use to create the integer literal here?

packages/core/src/transform/hash.ts
DuncanUszkay1: This is typescript not assemblyscript right? Sometimes its hard for me to tell, but I'm pretty sure that the transforms have to be written as actual TS.

If so I'd use something like this: https://www.npmjs.com/package/ts-md5

jtenner: @willemneal do you have any suggestions? I was trying to fit the hash of a string into a single 64 bit number.

jtenner: At @DuncanUszkay1 's suggestion, I decided to use djb2 algorithm. Please see here: http://www.cse.yorku.ca/~oz/hash.html.

willemneal: @DuncanUszkay1 @jtenner, The ultimate goal for transforms is to eventually be written in as, so not using ts dependencies isn't a bad idea.  However, there are already hash libraries in as and it would be easy to transition to them when the switch happens.  @jtenner also shouldn't you use BigInt if you want 64 bits?

jtenner: @willemneal how would I use a `BigInt` in the AST? What function call would I use?

jtenner: https://github.com/jtenner/as-pect/pull/254/files#diff-b9675e04d70b262d938410d8ce0cf0beR216

Please see here

PR #259, opened 2020-03-07
Remove csv-stringify from devDependencies in root package.json

9oelM:
csv-stringify should only exist as a dependency of `csv-reporter`


jtenner:
Thank you!

PR #261, opened 2020-03-09
Feature: Snapshots

jtenner:
This pull request comes with a lot of changes to every single package. 

## assembly package changes:

1. `expect(expected).toMatchSnapshot(name)` added to `Expectation<T>` members
2. Added `Reflect.toReflectedValue` parameter to indicate visual inspection vs snapshot inspection
3. Added `Expected.reportSnapshot(id, name)` api to report snapshots manually.

## CLI package changes:

1.  `--update` flag works just like `jest` to update calculated snapshots
2. `./test-filename.spec.ts` writes snapshots to `./__snapshots__/test-filename.spec.snap`
3. No way to currently work in an interactive mode. Snapshot updates are all or nothing (for now)

## core package changes

1. Added linked function to collect reported snapshots
2. added `snapshots` property for collected snapshots
3. added `expectedSnapshots` parameter and property to contain the loaded snapshots from the host
4. added `snapshotDiff` for snapshot diff calculations after `run()`

## snapshots package added:

1. Generic string snapshot comparisons
2. Added Snapshot class with a `Map` that contains the snapshots
3. Added SnapshotDiff class to contain the result of diffing two snapshots
4. uses `diff.diffLines` to calculate changes

TODO: 
- [x] add snapshot package
- [x] test snapshot output
- [x] add snapshots to core
- [x] report snapshots from assembly
- [x] report snapshots not for visual output
- [x] figure out why reporter isn't reporting snapshots in jest
- [x] add types
- [x] validate diff order
- [x] cleanup Summary and Verbose reporter output

Left for review:

- [x] validate necesity for overriding `__aspectStrictEquals()`

cc @DuncanUszkay1


packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -93,19 +94,23 @@ export class Reflect {
...
-        let displayValue = value.__aspectDisplayAs();
-        if (
-          !isInteger(displayValue) &&
-          !isFloat(displayValue) &&
-          !isManaged(displayValue)
-        ) {
-          ERROR(
-            "__aspectDisplayAs() function should return a managed type or a number",
-          );
+        if (forDisplay) {
```
DuncanUszkay1: Is this flag necessary? Feels like the test console output and the snapshot output should match

jtenner: That's a very good question. I think it's because it's named `__aspectDisplayAs()` vs `__aspectReflectAs()` and I wouldn't be opposed to renaming it. 

In the case of snapshots:

```ts
class A {
  a: i32 = 0;
  b: i32 = 1;
  c: i32 = 2;

  __aspectReportAs(): string {
    return "something static"; // does not vary by a, b and c
  }
}
```

If the snapshot validates the `string` output of `__aspectReportAs()`, the snapshot will not reflect the properties `a`, `b`, and `c`. 

I'm not arguing against your suggestion, I'm just stating what will happen, and users might get confused.

DuncanUszkay1: That's a good point- although I'd argue that if a user isn't using all the fields of the class in the `__aspectDisplayAs()` method then they aren't using it correctly anyway, since they won't be able to debug in the case of a test failure.

I guess the question is what's more confusing, having your test output on the console not match the snapshot or having your fields that were already omitted on the console not be used to test against the snapshot. Personally I'd go with option 1, but I imagine people will have different opinions

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -602,10 +623,20 @@ export class Reflect {
       let result = false;
       if (isNullable<T>()) {
         // @ts-ignore: __aspectStrictEquals is defined at this point, and left and right cannot be null
-        result = left!.__aspectStrictEquals(right!, stack, cache, [] as StaticArray<i64>);
+        result = left!.__aspectStrictEquals(
+          right!,
+          stack,
+          cache,
+          [] as StaticArray<i64>,
```
willemneal: Why is this a static array?

jtenner: - less heap allocations
- technically faster lookups
- concatenation is faster
- no need to call `.push()` and `.pop()`
- no need to change length of array

This is a good location to use `StaticArray` especially because we don't optimize the output

willemneal: I guess I didn't fully understand staticArray.  I thought it was like in rust where it has a static length known at compile time.  So it seemed weird to make it an empty one.

Is there a const length for it?

packages/core/__tests__/setup/SummaryReporterWrapper.ts
```diff
@@ -0,0 +1,77 @@
...
+        this.result += input;
+      },
+      reset() {
+        this.result = "";
+      },
+    };
+    this.stdout = this.writer;
+    this.stderr = this.writer;
+  }
+  // @ts-ignore
```
willemneal: Why the ignore?

jtenner: I know a better way to fix this. Just haven't gotten around to it.

jtenner: Thanks for the push. I fixed this.

jtenner:
I will cut this pull request with all the fixes just pulled in a version release tonight that relies on assemblyscript 0.9.3

Just need to fix merge conflicts

PR #267, opened 2020-03-13
Strict Equality for ArrayLike

jtenner:
@DuncanUszkay1 mentioned to me that ArrayLike values were not being compared properly in strict equality comparisons.

This PR aims to enable ArrayLike comparisons.

### DIFF
--- a/packages/assembly/assembly/internal/Reflect.ts
+++ b/packages/assembly/assembly/internal/Reflect.ts
@@ -563,36 +563,48 @@ export class Reflect {
           stack.pop();
           return result;
         }
-      }
 
-      // compile time array values should be compared over a for loop
-      if (left instanceof ArrayBufferView) {
-        // @ts-ignore: typesafe access to length
-        let aLength = left.length;
-        // @ts-ignore: typesafe access to length
-        let bLength = right.length;
+        // compile time array values should be compared over a for loop
+        if (isDefined(left.length)) {
+          // @ts-ignore: typesafe access to length
+          let aLength = left.length;
+          // @ts-ignore: typesafe access to length
+          let bLength = right.length;
 
-        // assert the lengths are good
-        if (aLength != bLength) return Reflect.FAILED_MATCH;
+          // assert the lengths are good
+          if (aLength != bLength) return Reflect.FAILED_MATCH;
 
-        // check each item
-        for (let i = 0; i < aLength; i++) {
-          let result = Reflect.equals(
-            // @ts-ignore: typesafe and runtime check safe array access
-            unchecked(left[i]),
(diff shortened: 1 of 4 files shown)


packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -563,36 +563,48 @@ export class Reflect {
...
-            unchecked(left[i]),
-            // @ts-ignore: typesafe and runtime check safe array access
-            unchecked(right[i]),
-            stack,
-            cache,
-          );
-          if (result == Reflect.FAILED_MATCH) return Reflect.FAILED_MATCH;
-        }
+        // compile time array values should be compared over a for loop
+        if (isDefined(left.length)) {
```
DuncanUszkay1: Are we assuming that any array with a length property is ArrayLike? It looks to me like  we could get to line 592 and fail- should we do something closer to the isArrayLike builtin in AS?
```typescript
var lengthField = this.lookupInSelf("length");
     return lengthField !== null && (
       lengthField.kind == ElementKind.FIELD ||
       (
        lengthField.kind == ElementKind.PROPERTY &&
         (<Property>lengthField).getterInstance !== null // TODO: resolve & check type?
       )
     ) && (
       this.lookupOverload(OperatorKind.INDEXED_GET) !== null ||
       this.lookupOverload(OperatorKind.UNCHECKED_INDEXED_GET) !== null
```
Or is this sufficient on its own?

jtenner: I believe we check the same things the compiler does already, and the tests seem to pass. It appears to be sufficient.

jtenner: Is it okay to ask for your Cstr implementation so I can write some tests against it for you to verify?

DuncanUszkay1: The tests show that this works for array like classes, I was referring to non-array like classes which happen have a length field:
```typescript
class Race {
  length: i32, // length of the race in miles
  participants: ...
}
```
This would be interpreted as an array and fail on line 592 right? Feels like we need this check up front to prevent that:
```typescript
isDefined(unchecked(left[0])) || isDefined(left[i])
```

I don't think there's any need to send our implementation, if our implementation doesn't work with this then we should probably change our implementation.

jtenner: A `isDefined(left[0])` already validates an index signature before this `if` statement is reached. It means at this point, there is an index signature and a length, essentially performing the equalivalent of `isArrayLike(left)`.

jtenner:
We should snapshot the output of this expectation in `packages/core/assembly/jest-fail.ts`

PR #269, opened 2020-03-16
v0.10 migration

9oelM:


packages/assembly/assembly/internal/Expectation.ts
```diff
@@ -53,11 +53,13 @@ export class Expectation<T> {
     Actual.report(actual);
 
     if (isReference<T>() && !isFunction<T>()) {
-      if (!negated
-        && changetype<usize>(actual) !== 0
```
willemneal: Shouldn't the lines below shift to be inline with the first.  I'm not a fan of a line`if (`.

If we were to try improve the style here I propose creating a simple boolean function that describes what is being checked.

packages/cli/src/help.ts
```diff
@@ -10,6 +10,7 @@ export function help() {
   {bold.blueBright SYNTAX}
     {bold.green asp} --init                          Create a test config, an assembly/__tests__ folder and exit.
     {bold.green asp} -i
+    {bold.green asp} --init-wasi                     Create a test config for use with wasi, an assembly/__tests__ folder and exit.
```
willemneal: Why isn't this an argument for the init command.   In `packages/cli/src/util/CommandLineArg.ts` you'll find

```ts
  init: {
    description: "Create a test config, an assembly/__tests__ folder and exit.",
    type: "b",
    alias: { name: "i" },
    value: false,
  },
```
just change this to

```ts
  init: {
    description: "Create a test config, an assembly/__tests__ folder and exit.",
    type: "str",
    alias: { name: "i" },
    value: "default",
  },
```
Then users can pass `--init wasi` and another idea would be allow for projects to create their own init file for project specific details.

packages/cli/src/init.ts
```diff
@@ -51,12 +51,16 @@ export function init() {
       createWriteStream(typesFile, "utf-8"),
     );
   }
+
+  const configFileName: string = wasi
+    ? "as-pect.wasi.config.js"
```
willemneal: Why isn't this just part of the normal config file?

packages/core/__tests__/setup/createTestGroupFilterModule.ts
```diff
@@ -2,13 +2,14 @@ import { main } from "assemblyscript/cli/asc";
...
   ASUtil,
+  ResultObject,
 } from "assemblyscript/lib/loader";
 import { TestContext } from "../../src/test/TestContext";
 import { EmptyReporter } from "../../src/reporter/EmptyReporter";
 import { IAspectExports } from "../../src/util/IAspectExports";
 
 interface ICreateModuleCallbackResult {
-  wasm: ASUtil & IAspectExports;
+  wasm: ResultObject & { exports: ASUtil & IAspectExports };
```
willemneal: Since this is use frequently enough perhaps we should use a new type alias.

packages/cli/src/help.ts
jtenner: Yeah. This particular solution won't be good moving forward, so I'm opting to change the default configuration to have an optional `wasi` parameter which allows users to specify if they want `wasi` instantiation.

packages/cli/src/init.ts
jtenner: This feature is going to be removed

packages/core/__tests__/setup/createTestGroupFilterModule.ts
jtenner: These types are going to be exposed via the loader once assemblyscript updates versions, so I'm going to keep this as it is, if that's okay.

packages/cli/src/init.ts
jtenner: Exactly.

packages/assembly/assembly/internal/Expectation.ts
jtenner: This is what prettier wants, so I'm going to mark this as resolved. The code style is rather strange, however.

PR #270, opened 2020-03-16
fix nullable reflected references

jtenner:
We should assert that `left` and `right` values, when compared, are not nullable by the time the flow hits the `instanceof` checks.


packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -454,7 +458,7 @@ export class Reflect {
       if (isDefined(left[0])) {
         // test for safe indexof usage
         // set match
-        if (left instanceof Set) {
+        if (left! instanceof Set) {
```
DuncanUszkay1: why do we need to assert here again? Shouldn't line 412 already guarantee left is not nullable?

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -397,23 +397,22 @@ export class Reflect {
...
-      return Reflect.FAILED_MATCH;
-    }
-
-    // if it's possible for T to be null
-    if (isNullable<T>()) {
-      // mutual exclusion null
-      if (
-        i32(changetype<usize>(left) == 0) ^ i32(changetype<usize>(right) == 0)
-      )
+    if (isReference(left)) {
```
willemneal: Why not `isReference<T>()`?

packages/assembly/assembly/internal/Reflect.ts
jtenner: Yeah that's part of the flow error I've found. Not sure how to fix it.

packages/assembly/assembly/internal/Reflect.ts
jtenner: I don't mind relying on type inference here.

willemneal: Ah well since you did, `isNullable<T>()` above, we should choose one or the other for consistency.

jtenner: fair enough. I'll just change it.

PR #274, opened 2020-03-20
WASI support

### LINKED ISSUE #246: How to test methods that make WASI calls?
Given a file `sample.ts`:

```typescript
import { Console } from "as-wasi";

export function sample(): void {
    Console.log("Foo");
}
```

and a test file `sample.spec.ts`:

```typescript
import { sample } from "../src/sample";

describe("sample", () => {
  it("Should verify something", () => {
    sample();
  });
});
```

I get:
```
TypeError: WebAssembly.Instance(): Import #1 module="wasi_unstable" error: module is not an object or function
    at instantiateSync (/Users/jeffreycharles/scripts/feb27/node_modules/assemblyscript/lib/loader/index.js:307:5)
    at runBinary (/Users/jeffreycharles/scripts/feb27/node_modules/@as-pect/cli/lib/run.js:366:28)
    at /Users/jeffreycharles/scripts/feb27/node_modules/@as-pect/cli/lib/run.js:482:46
    at Object.main (/Users/jeffreycharles/scripts/feb27/node_modules/assemblyscript/cli/asc.js:775:10)
    at /Users/jeffreycharles/scripts/feb27/node_modules/@as-pect/cli/lib/run.js:433:21
    at Array.forEach (<anonymous>)
    at run (/Users/jeffreycharles/scripts/feb27/node_modules/@as-pect/cli/lib/run.js:431:40)
    at Object.asp (/Users/jeffreycharles/scripts/feb27/node_modules/@as-pect/cli/lib/index.js:72:13)
    at Object.<anonymous> (/Users/jeffreycharles/scripts/feb27/node_modules/@as-pect/cli/bin/asp:2:28)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
```

How can I configure `as-pect` so that the `Console.log` call works in the test? Ideally that it prints `Foo` to stdout, less ideally, that the test executes successfully.

jtenner:
Fixes #246 

This should be ready for review and testing. 

CC @willemneal

### DIFF
--- a/packages/cli/src/util/AspectConfiguration.ts
+++ b/packages/cli/src/util/AspectConfiguration.ts
@@ -66,16 +66,16 @@ export interface AspectConfiguration {
   /** A wasi configuration. */
   wasi?: {
     /** An array of strings that the WebAssembly application will see as command line arguments. The first argument is the virtual path to the WASI command itself. Default: []. */
-    args: string[];
+    args?: string[];
     /** An object similar to process.env that the WebAssembly application will see as its environment. Default: {}. */
-    env: {
+    env?: {
       [envVar: string]: string; // value
     };
     /** This object represents the WebAssembly application's sandbox directory structure. The string keys of preopens are treated as directories within the sandbox. The corresponding values in preopens are the real paths to those directories on the host machine. */
-    preopens: {
+    preopens?: {
       [virtualPath: string]: string;
     };
     /** By default, WASI applications terminate the Node.js process via the __wasi_proc_exit() function. Setting this option to true causes wasi.start() to return the exit code rather than terminate the process. Default: false */
-    returnOnExit: boolean;
+    returnOnExit?: boolean;
   };
 }
--- a/packages/cli/src/util/AspectConfiguration.ts
+++ b/packages/cli/src/util/AspectConfiguration.ts
@@ -3,7 +3,7 @@ import { IReporter } from "@as-pect/core";
 /**
  * This is the shape of the compiler flags.
  */
-export interface ICompilerFlags {
+export interface CompilerFlags {
   [flag: string]: string[];
 }
 
@@ -11,7 +11,7 @@ export interface ICompilerFlags {
  * This is an interface describing the shape of an exported configuration for the
  * `as-pect.config.js` file. An empty object should be a valid `as-pect` configuration.
  */
-export interface IConfiguration {
+export interface AspectConfiguration {
   [key: string]: any;
   /**
    * A set of globs that denote files that must be used for testing.
@@ -24,7 +24,7 @@ export interface IConfiguration {
   /**
    * The compiler flags needed for this test suite. Do not forget that a binary file must be output.
    */
-  flags?: ICompilerFlags;
+  flags?: CompilerFlags;
   /**
    * A set of regular expressions that are tested against the file names. If they match, the
    * files will be discluded.
@@ -68,7 +68,9 @@ export interface IConfiguration {
     /** An array of strings that the WebAssembly application will see as command line arguments. The first argument is the virtual path to the WASI command itself. Default: []. */
     args: string[];
     /** An object similar to process.env that the WebAssembly application will see as its environment. Default: {}. */
-    env: any;
+    env: {
+      [envVar: string]: string; // value
+    };
     /** This object represents the WebAssembly application's sandbox directory structure. The string keys of preopens are treated as directories within the sandbox. The corresponding values in preopens are the real paths to those directories on the host machine. */
     preopens: {
       [virtualPath: string]: string;
--- a/packages/cli/src/util/AspectConfiguration.ts
+++ b/packages/cli/src/util/AspectConfiguration.ts
@@ -72,7 +72,7 @@ export interface AspectConfiguration {
       [envVar: string]: string; // value
     };
     /** This object represents the WebAssembly application's sandbox directory structure. The string keys of preopens are treated as directories within the sandbox. The corresponding values in preopens are the real paths to those directories on the host machine. */
-    preopens?: {
+    preopens: {
       [virtualPath: string]: string;
     };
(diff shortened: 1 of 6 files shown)


packages/cli/src/util/IConfiguration.ts
```diff
@@ -63,4 +63,17 @@ export interface IConfiguration {
    * WASM Memory max size in pages 64kb. Should be positive. Default is disabled or -1.
    */
   memoryMax?: number;
+  /** A wasi configuration. */
+  wasi?:  {
+    /** An array of strings that the WebAssembly application will see as command line arguments. The first argument is the virtual path to the WASI command itself. Default: []. */
+    args: string[];
+    /** An object similar to process.env that the WebAssembly application will see as its environment. Default: {}. */
+    env: any;
```
willemneal: this will be string indexed correct?  If so, it should be similar to how preopens is defined below.

packages/core/src/test/TestContext.ts
```diff
@@ -57,6 +57,8 @@ export interface ITestContextParameters {
   reporter: IReporter;
   /** The expected snapshot output. */
   snapshots?: Snapshot;
+  /** Wasi parameters. */
+  wasi?: any;
```
willemneal: Are there not types provide by node for wasi?

packages/core/src/test/TestContext.ts
```diff
@@ -237,141 +256,149 @@ export class TestContext {
...
       this.addResult(node, false);
       this.reporter.onExit(this, node);
       return;
     }
-
-    // We now have the responsibility to run each beforeAll callback before traversing children
-    if (!this.runFunctions(node.beforeAll)) {
+  } else {
+    // gather all the tests and groups, validate program state at this level
+    const success = this.tryCall(node.callback) === 1;
```
willemneal: shouldn't node.performanceStart be set here and above in the `negated case`?

packages/core/src/test/TestContext.ts
jtenner: @types/node does not have wasi types and I'm perfectly okay with `any` right now as support. We can improve this later.

packages/core/src/test/TestContext.ts
jtenner: node.performanceStart is already set at this point. Yes.

packages/cli/src/util/IConfiguration.ts
jtenner: I'm not trying to have perfect types for this wasi support release. I suppose we should improve these types when `@types/node` updates. Is that an okay compromise?

packages/core/src/test/TestContext.ts
willemneal: No I mean that it seems that you are recording things that aren't the execution of the node.

packages/cli/src/util/IConfiguration.ts
willemneal: Okay then perhaps add a `TODO` comment?

jtenner: Updated. Please advise.

packages/core/src/test/TestContext.ts
jtenner: Yeah. You're probably right. Can we file an issue and fix this in another pull request?

packages/core/src/test/TestContext.ts
```diff
@@ -540,6 +567,11 @@ export class TestContext {
     /** Override trace completely. */
     finalImports.env.trace = this.trace.bind(this);
 
+    /** Add support for wasi if requested. */
+    if (this.wasi) {
+      finalImports.wasi_snapshot_preview1 = this.wasi.wasiImports;
```
andrewhassan:
```suggestion
      finalImports.wasi_snapshot_preview1 = this.wasi.wasiImport;
```

I was just testing this branch with Node 13.11.0 and it was breaking on this line. Looks like this should be `wasiImport`, not `wasiImports`.

jtenner: Thanks for the help. I'm not entirely sure how this should be tested, to be honest. What kind of AssemblyScript code would be appropriate to validate that WASI is functioning?

jtenner:
Docs got updated here: https://app.gitbook.com/@tenner-joshua/s/as-pect/cli-configuration

packages/core/src/test/TestContext.ts
andrewhassan: No problem! I actually don't have any formal tests or anything. I just created a new test that imports `Console` from `as-wasi`, and then manually checked to see that `Console.log('test')` worked.

packages/cli/src/util/AspectConfiguration.ts
```diff
@@ -63,4 +63,19 @@ export interface IConfiguration {
    * WASM Memory max size in pages 64kb. Should be positive. Default is disabled or -1.
    */
   memoryMax?: number;
+  /** A wasi configuration. */
+  wasi?: {
```
andrewhassan: In the case that a user wants to use the default `wasi` config, with this interface, they're forced to actually set the default values (and still be typed correctly). I think it might make sense if each key with a default value was optionally typed like:
```
wasi?: {
  args?: string[];
  
  env?: {
    [envVar: string]: string; // value
  };

  preopens: {
    [virtualPath: string]: string;
  };

  returnOnExit?: boolean;
};
```

jtenner: Thanks for the tip. I think you're correct.

andrewhassan: No problem! Great job on as-pect btw. It's awesome!

packages/core/src/test/TestContext.ts
jtenner: I filed an issue for tightening up the node start and end times to be more accurate. (We are talking less than millisecond timing accuracy, but could technically be made better.

PR #277, opened 2020-03-21
Feature: SIMD

jtenner:
This pull request seeks to support SIMD directly.


jtenner:
@willemneal it bombs out saying: invalid local. Any ideas?

packages/assembly/assembly/internal/Reflect.ts
```diff
@@ -61,8 +61,33 @@ export class Reflect {
     value: T,
     seen: Map<usize, i32> = new Map<usize, i32>(),
   ): i32 {
-    // if T is a reference
-    if (isReference<T>()) {
+
+    if (ASC_FEATURE_SIMD && value instanceof v128) {
```
jtenner: This actually doesn't work because `v128` cannot be resolved here if ASC_FEATURE_SIMD is disabled. Need to refactor this.

jtenner:
I will attempt to try this at some point. Now is not the time.

PR #279, opened 2020-03-24
Fix default reporter

willemneal:
Setup summary reporter correctly so that it has all writeable streams.


packages/cli/src/util/collectReporter.ts
```diff
@@ -90,15 +75,26 @@ export function collectReporter(cliOptions: Options): IReporter {
...
+        : {},
+    );
+    reporter.stdout = process.stdout;
+    reporter.stderr = process.stderr;
+    reporters.push(reporter);
     process.stdout.write(
       chalk`{bgWhite.black [Log]} Using {yellow SummaryReporter}\n`,
     );
-    return new SummaryReporter({
-      enableLogging: true,
```
jtenner: We need to make sure logging is enabled by default

PR #285, opened 2020-04-28
implement static heap string cache

jtenner:
Feature: Add static string level caching. Useful for speeding up usage of the reflection api, when it's heavily used.

Changes:
- Added a test for calculated strings to increase code coverage
- 100% code coverage
- Exports the heap base now
- `getString()` is now protected
 
Thoughts:

It's hard to read the snapshot changes because when the stack traces change, it creates a very large change to the snapshots. This is something we should improve in a later pull request.

Perhaps each of these Reflected value snapshots should just be stringified?

### DIFF
--- a/packages/core/src/test/TestContext.ts
+++ b/packages/core/src/test/TestContext.ts
@@ -207,8 +207,6 @@ export class TestContext {
     // set the wasm
     this.wasm = wasm;
 
-    this.__heap_base = wasm.__heap_base.value;
-
     // start by visiting the root node
     this.visit(this.rootNode);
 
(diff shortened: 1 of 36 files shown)


jtenner:
Per @willemneal's idea, I was able to generalize this pattern way better. I did confirm the `cachedStrings.delete()` method is called on an allocated string (the calculated one in `log.ts` for the core tests.)

packages/core/src/test/TestContext.ts
```diff
@@ -792,6 +792,9 @@ export class TestContext {
       this.blocks.delete(block);
     }
 
+    // remove any cached strings at this pointer
+    this.cachedStrings.delete(block + 16);
```
willemneal, line 796: What happens on a delete to a non-existent key?  Since it could be the case that you never exported the string, but deallocated it.

jtenner: `this.cachedStrings.delete(block + 16)` simply returns false if it doesn't delete anything.

PR #288, opened 2020-05-10
remove isDefined for unchecked, code coverage problem, long.js usage

jtenner:
`unchecked()` is always safe here, since it always defaults to regular checked gets if the operator doesn't exist. I also remove a few default parameters because they are handled in AssemblyScript.

### DIFF
--- a/packages/assembly/assembly/__tests__/toBe.spec.wat
+++ b/packages/assembly/assembly/__tests__/toBe.spec.wat
@@ -5850,7 +5850,7 @@
   else
    i32.const 0
    i32.const 2912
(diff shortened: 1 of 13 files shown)


jtenner:
Okay. I lied.

This pull request is going to be generalized improvements. It was worth writing a helper function to help remove all the stack traces manually. We can improve this later, and it makes snapshot output way less verbose for now.

jtenner:
Think we should be good for an official release to 3.1.6?

willemneal:
I'd think so given the scope of changes.

PR #291, opened 2020-05-11
What if the assert function lived in TypeScript instead...?

jtenner:
Less expression compiling. Gives a mild speed boost.


jtenner:
This would be fascinating to implement. And again, pulling stuff out into the host will reduce compilation times. However, this is not the primary bottleneck of as-pect at the moment. Closing because there's no reason to optimize this when there are plenty of other perfectly reasonable optimizations to implement.

PR #293, opened 2020-05-15
10 migration

jtenner:
Now that version 10 has released, we should probably release a version of as-pect with wasi support and the new loader support.

All of this is quick and dirty so we can get things started.

@willemneal and @DuncanUszkay1, any chance I could get some help testing this locally?

I am moving to get this pulled as soon as possible, with the promise that we can *AND SHOULD* refactor this a bit later.


packages/assembly/assembly/__tests__/RTrace.spec.wat
```diff
@@ -1587,17 +1655,28 @@
...
   call $assembly/internal/Expectation/Expectation<i32>#constructor
  )
- (func $assembly/internal/Reflect/Reflect.toReflectedValue<i32> (; 35 ;) (param $0 i32) (param $1 i32) (result i32)
+ (func $assembly/internal/Reflect/Reflect.toReflectedValue<i32> (param $0 i32) (param $1 i32) (result i32)
   (local $2 i32)
   (local $3 i32)
   local.get $1
   call $~lib/rt/pure/__retain
   local.set $1
+  i32.const 0
```
jtenner, line 1669: This is all very unfortunate.

packages/assembly/assembly/__tests__/RTrace.spec.wat
```diff
@@ -2076,6 +2103,13 @@
   local.get $1
   local.get $4
   call $assembly/internal/Expected/Expected.report<i32>
+  i32.const 1
```
jtenner, line 2106: Another unfortunate change.

packages/assembly/assembly/__tests__/RTrace.spec.wat
```diff
@@ -2183,18 +2217,29 @@
...
  )
- (func $assembly/internal/Reflect/Reflect.toReflectedValue<u32> (; 55 ;) (param $0 i32) (param $1 i32) (result i32)
+ (func $assembly/internal/Reflect/Reflect.toReflectedValue<u32> (param $0 i32) (param $1 i32) (result i32)
   (local $2 i32)
   (local $3 i32)
   local.get $1
   call $~lib/rt/pure/__retain
   local.set $1
   i32.const 0
+  drop
```
jtenner, line 2232: Seems a lot of these happen. Looks like toReflectedValue might benefit from some refactoring?

packages/core/src/test/TestContext.ts
```diff
@@ -1472,4 +1495,17 @@ export class TestContext {
       this.reflectedValueCache[reflectedValueID].stringify(stringifyOptions),
     );
   }
+
+  private wasiStart(): void {
```
jtenner, line 1499: @willemneal I know this looks awful, but the native `wasi.start(instance)` method was definitely not working. Here, what I had to do was obtain the symbols in the wasi instance, and set the memory manually, and run the `_start()` function manually.

jtenner: I removed this in favor of mocking it in jest where the problem lies.

jtenner:
@willemneal this is high priority. Can you help me validate that wasi support is working?

willemneal:
@jtenner Can confirm that your migration PR works locally.  However, you need to have node >= 13.

jtenner:
I will make sure that everything is opt in. When the wasi option is passed, that's when it requires wasi, so older versions of node should be safe.

PR #294, opened 2020-05-17
Major CLI Rewrite [Breaking Change]

jtenner:
This pull request has a huge intent to refactor the way the entire cli works, leaving the TestContext classes in the core module untouched. This change will be a breaking change only to the cli and as-pect.config.js interface, and thus, users of aspect will have to change the way they interface with as-pect in the console in order to upgrade.

This breaking change comes along with a `--compiler` flag, which lets the user directly specify the compiler location, so that it is always used, even for the core transforms. This should fix a long standing problem where @willemneal would have to do lots of hacky things to assert that only one version of binaryen was loaded at any given time.

- [x] Switch to AssemblyScript cli parser
- [x] Implement options for cli
- [x] redefine the shape of the as-pect.config.js file
- [x] implement a `AspectRunContext` object that will contain an "effective" configuration for the entire run
- [x] create and define a set of tasks that will exist in `packages/cli/src/tasks`
- [x] use promises so that each run step of the cli is perfectly clear and maintainable
- [x] implement printing pre-defined error codes using a regex and the cli run context
- [ ] implement a `TestRunContext` for each entry module

Notes:

When running the tests in 3.2 and a module causes a panic, the error is not handled gracefully with a catch, and any files that would have been written to the file system become truncated because the process exits code one immediately. This includes snapshots, wat files, and wasm files etc. This pull request aims to solve this problem by implementing a top level `catch()` in `run.ts` so that the process can `await` all of the file write promises before exiting code `1` which is expected behavior.

With this pull request, the compiler resolution algorithm will also be rewritten, but it introduces new problems to be fixed. In order to resolve the `assemblyscript/cli/util/options` module to parse our cli arguments, we have to resolve the compiler location first. Furthermore, we can't even display the help screen using the built-in `help()` function without it. This means we must check for the `--compiler` and `-C` flag by parsing `argv` manually. This leaves us with a potential for the compiler to remain unresolved before loading the as-pect config. Here are the proposed steps for configuration and cli resolution:

1. If the compiler can be resolved with the cli args, immediately parse the cli arguments and continue to resolve the configuration.
2. If the compiler cannot be resolved at this step, we must manually resolve the assemblyscript configuration without the use of the `parse()` function for the cli arguments. This means inspecting the `--config` or `-c` flag manually and resolve the as-pect.config (which should essentially be optional)
3. Once we have the as-pect.config (or not!), we need to finally resolve the compiler if it hasn't been resolved yet by inspecting the `"compiler"` property in the configuration. If it doesn't exist, then default to `"assemblyscript"` and resolve the compiler.
4. If no compiler can be found, throw an as-pect error and exit code 1

We will use a new `ICLIRunContext` which should contain (at least) the following things:
```ts
interface ICLIRunContext {
  stdout: IWritable;
  stderr: IWritable;
  exit(code: number): void;
  config: IAspectConfiguration;
  cliResult: Result; // from parse function
  filePromises: Promise<void>[]; // to keep track of each promise that gets created
  exitCode: number; // once we start using the compiler, set this exitCode instead
}
```

This pr is a long time coming.


jtenner:
@willemneal any comments so far? @DuncanUszkay1 and @9oelM might be able to chime in too.

packages/cli/src/AspectRunContext.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+
+export interface IOptionalRunContextProps {
+  /** The assemblyscript help text generation function. */
+  help: typeof import("assemblyscript/cli/util/options").help;
+  /** The assemblyscript cli parse function. */
+  parse: typeof import("assemblyscript/cli/util/options").parse;
+  /** The result of calling the assemblyscript cli parse function. */
+  parseResult: import("assemblyscript/cli/util/options").Result;
```
willemneal, line 29: Wouldn't this be null until the parse function is called?

jtenner: Yes. That's why I use the `Partial<T>` type with the optional types.

packages/cli/src/AspectRunContext.ts
```diff
@@ -0,0 +1,48 @@
...
+  argv: string[];
+  /** A set of file promises that must be resolved before the application exits. */
+  files: Promise<void>[];
+  /** A stripped down version of a process that contains an exit function, stdout and stderr. */
+  process: IProcessConfiguration;
+  /** The exit code used to exit the program. It starts as 0, and can only be set to 1 otherwise. */
+  exitCode: number;
+}
+
+export type AspectRunContext = Partial<IOptionalRunContextProps> &
```
jtenner, line 47: @willemneal see here.

willemneal:
Looks good so far.  One idea I have is to separate configuration with plugins (e.g. reports, etc).  This way the config has parity with the CLI.

jtenner:
Okay. There may be some issues with the initial steps. We will have to discover them as we continue development. For now, the steps seems to be good enough. I'm going to move onto collecting and running test modules.

jtenner:
One of the things that I note about this pull request is that it won't take advantage of asconfig.

I believe another entire rewrite of how the cli works is going to be necessary. 

1. We need an as-pect.asconfig.json which configures the cli options for a single compilation (default cli run does a single compilation `asc --config as-pect.asconfig.json`) 
2. As-pect.imports.js file in root directory for module imports
3. as-pect.wasi.js for wasi configuration
4. No custom reporters? (this doesn't seem like a popular feature) I want convention over configuration and ease of use.

The only problem I see with this new approach is "how do we define entry points without globs?" perhaps that might be better done in asconfig?

jtenner:
I am closing this because it didn't feel right when I designed it. Current aspect works as intended.

Issue #282, opened 2020-04-22
StrictEquality print depth


DuncanUszkay1:
Right now when you compare two deeply nested objects, the logged result is truncated. Consider the following example:
```typescript
class LayerA {
  parent: LayerB
}
class LayerB {
  parent: LayerC
}
class LayerC {
  parent: LayerD
}
class LayerD {
  foo: i32
}
describe("example", () => {
  it("should print as expected", () => {
    let nested: LayerA = {
      parent: {
        parent: {
          parent: {
            foo: 7
          }
        }
      }
    }
    let wrongNested: LayerA = {
      parent: {
        parent: {
          parent: {
            foo: 20
          }
        }
      }
    }
    expect<LayerA>(nested).toStrictEqual(wrongNested);
  });
});
```
outputs 
```
┃ ❌ test/script.spec.ts Pass: 0 / 1 Todo: 0 Time: 6.548ms
┃   Failed: example
┃     ❌ should print as expected -
┃       [Actual]  : LayerA {
┃       parent: LayerB {
┃           parent: LayerC {
┃               parent: [LayerD]
┃           }
┃       }
┃   }
┃       [Expected]: LayerA {
┃       parent: LayerB {
┃           parent: LayerC {
┃               parent: [LayerD]
┃           }
┃       }
┃   }
```
Which hides the foo field.

jtenner:
I'm not sure how to fix this, and traversal depth is a feature by design to prevent deeply nested objects from flooding stdout.

I'm open to suggestions, however.

DuncanUszkay1:
Would it make sense to print out the full object (barring circular references) in verbose mode? If I'm comparing two large objects in a test in verbose mode, and I get a lot of output, is that a problem?

We would have to deal with circular references in this case, though given that there's already a similar algorithm implemented to do the comparison it seems that should be doable.

Thoughts?

jtenner:
Circular references are already handled in `Reflection.toReflectedValue()`. But you make a very good point. I will be sure to include this in the breaking changes for 4.0.

PR #299, opened 2020-05-20
Increase verbose reporter expand depth

DuncanUszkay1:
Issue: https://github.com/jtenner/as-pect/issues/282

When printing a deeply nested object in verbose mode, the input was getting cut off fairly quickly. I've increased the expand level for the verbose reporter from the default of 3 to 10. This allows verbose mode to be truly verbose.


jtenner:
Wow. I had no idea that property wasn't being respected. Thanks for the pull request. 10 seems like a reasonable value to include. 

There is a small chance this might be an obsolete change because there might be some breaking changes to reporters, however, as a fix for the `3.2.x` branch, this is great.

I just want to think about it for a day or so before I pull it.

jtenner:
Looks like snapshots weren't updated. Is travis working?

PR #302, opened 2020-05-30
support compiler with closure implemented

jtenner:
Some of the TypeNode functions were refactored so most of the work in this pull request was dealing with the compiler api in the core/transform functions.

I added a simple closure test to assert that everything is working, and it does.... albeit locally. 

Notes:

1. as-pect supports the latest stable version of the compiler, so this change is just a work in progress and is subject to change
2. we need to replace the `"assemblyscript": "file:../assemblyscript"` entry because I had to bootstrap everything locally
3. There's a nullable branch change there, because the tests became pedantic and I wanted to fix it

@willemneal you can feel free to review this change, even though there doesn't appear to be anything offensive.


jtenner:
Closing this. There's no reason to support this yet.

PR #309, opened 2020-06-14
v 0.11 migration

jtenner:


packages/core/src/util/wasmTools.ts
```diff
@@ -30,10 +30,12 @@ export class WasmBuffer {
...
     const name_len = this.readVaruint(off);
     this.off += name_len;
-    //@ts-ignore
-    return String.fromCharCodes(
-      this.u8array.slice(this.off - name_len, this.off),
-    );
+    const codes = this.u8array.slice(this.off - name_len, this.off);
+    let result = "";
+    for (let i = 0; i < codes.length; i++) {
+      result += String.fromCharCode(codes[i]);
```
willemneal, line 36: I just tested this out and it turns out this is the fasted way to build the string!
https://jsperf.com/concatenation-vs-map-and-join

jtenner: Yeah. I had to fix this because String.fromCharCodes was throwing an error, actually.

PR #316, opened 2020-10-26
AssemblyScript 17 upgrade

jtenner:
@willemneal Please review. I think we have some issues with the assembly package.

I also included a small fix to the snapshot parser.


packages/cli/src/run.ts
```diff
@@ -95,7 +95,12 @@ export function run(cliOptions: Options, compilerArgs: string[]): void {
     }
 
     /** Next, collect the loader, and assert it has an instantiateSync method. */
-    let loader = require(path.join(assemblyScriptFolder, "lib", "loader"));
+    let loader;
+    try {
+      loader = require(path.join(assemblyScriptFolder, "lib", "loader"));
+    } catch (ex) {
+      loader = import(path.join(assemblyScriptFolder, "lib", "loader"));
```
jtenner, line 102: Hey I thought `import()` returns a promise..?

jtenner (commented):
Honestly, these changes all look good. I'll get around to checking these all myself soon.

packages/cli/src/run.ts
willemneal: Ah you're right. Then I don't know why this fixed it and why it type checked fine.

willemneal: Really I should just to `require(..."/umd")`

packages/core/__tests__/__snapshots__/RTrace.spec.ts.snap
```diff
@@ -234,14 +234,36 @@ exports[`RTrace output Overall Statistics: RTrace testserrors 1`] = `Array []`;
...
 exports[`RTrace output Overall Statistics: a group reference counting error rtraceStart 1`] = `7`;
 
-exports[`RTrace output Overall Statistics: a group reference counting errorerrors 1`] = `Array []`;
-
-exports[`RTrace output Overall Statistics: allocationCount 1`] = `60`;
-
-exports[`RTrace output Overall Statistics: errors 1`] = `Array []`;
-
-exports[`RTrace output Overall Statistics: freeCount 1`] = `50`;
+exports[`RTrace output Overall Statistics: a group reference counting errorerrors 1`] = `
```
jtenner, line 241: We should take a look at this. What is this?

@willemneal

packages/core/src/test/TestContext.ts
```diff
@@ -1499,4 +1501,8 @@ export class TestContext {
       this.reflectedValueCache[reflectedValueID].stringify(stringifyOptions),
     );
   }
+
+  onresize(_block: number, _size: number): void {
```
jtenner, line 1505: The next pull request needs to implement the `RTrace` class. We bump the current version to 5.0, and then implement RTrace support in `5.1`.

packages/snapshots/src/parser/grammar.ne
```diff
@@ -1,12 +1,14 @@
...
 
-start -> _ lines _ {% d => {
+start -> _ (lines _):? {% d => {
   const results = d[1];
   const map = new Map<string, string>();
-  for (let i = 0; i < results.length; i++) {
-    const [key, value] = results[i];
-    if (map.has(key)) throw new Error("Invalid snapshot.");
-    map.set(key, value);
+  if (results) {
```
jtenner, line 6: This bug fix helps load empty snapshots. A looming bug.

jtenner:
closed in favor of #319

PR #319, opened 2020-11-24
Rtrace upgrade

jtenner:
@willemneal Got any ideas here?


willemneal (approved):

jtenner:
It looks like a rogue release, but I actually believe that the problem is a lack of `__retain` on a specific array.

```wat
(func $assembly/jest-log/testLogFunction
  (local $0 i32)
  (local $1 i32)
  (local $2 i32)
  i32.const 0
  i32.const 0
  call $~lib/array/Array<i32>#constructor
  local.set $0
  i32.const 0
  local.set $1
  loop $for-loop|0
   local.get $1
   i32.const 20
   i32.lt_s
   local.set $2
   local.get $2
   if
    local.get $0
    local.get $1
    local.get $1
    call $~lib/array/Array<i32>#__set
    local.get $1
    i32.const 1
    i32.add
    local.set $1
    br $for-loop|0
   end
  end
  local.get $0
  call $../assembly/assembly/internal/log/log<~lib/array/Array<i32>>
  local.get $0
  call $~lib/rt/pure/__release
 )
```

This is generated by this code snippet:

```ts
function testLogFunction(): void {
  let lotsOfValues = new Array<i32>();
  for (let i = 0; i < 20; i++) lotsOfValues[i] = i; // for expanded values
  log(lotsOfValues);
```

jtenner:
Code coverage fails, but I can reproduce local code coverage checks are at 100% of the lines we assert need to be covered.

Pulling now.

PR #327, opened 2021-02-10
As18 upgrade

jtenner:


packages/cli/src/run.ts
```diff
@@ -611,7 +595,10 @@ export function run(cliOptions: Options, compilerArgs: string[]): void {
       }
 
       Promise.all(filePromises).then(() => {
-        if (failed) process.exit(1);
+        if (failed) {
+          console.log(errors);
```
willemneal, line 599:
```suggestion
          console.error(errors);
```

willemneal (approved):
LGTM just one nit!

packages/cli/src/run.ts
jtenner: Shouldn't we remove this altogether? Let's fix this right now. 

What's the best way to report these errors to stderr? Is that enough?

willemneal: I think it's enough.  Though perhaps each error should be sent to console.error.  So `errors.map(console.error)`? Perhaps that prints better?

Issue #328, opened 2021-02-10
update for AssemblyScript 0.18.*?


bcolloran:
the instructions for installing as-pect at https://tenner-joshua.gitbook.io/as-pect/getting-started (using assemblyscript nightly) result in:

```
$ npm install --save-dev @as-pect/cli
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! Found: assemblyscript@0.10.0-nightly.20200611
npm ERR! node_modules/assemblyscript
npm ERR!   dev assemblyscript@"0.10.0-nightly.20200611" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! dev @as-pect/cli@"*" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: assemblyscript@0.17.14
npm ERR! node_modules/assemblyscript
npm ERR!   peer assemblyscript@"^0.17.10" from @as-pect/cli@5.0.1
npm ERR!   node_modules/@as-pect/cli
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

Perhaps the version numbers in package.json need to be bumped? (If as-pect is compatible with the latest versions of course)

bcolloran:
More information:

Unfortunately, it appears that there is genuine incompatibility between 0.17.* and 0.18.*, and more than a version bump will be required -- after installing as-pect according to the quick start instructions, but with `npm install --legacy-peer-deps --save-dev @as-pect/cli` and trying to the example tests in `example.spec.ts`, I'm encountering:

```
[Error] There was a compilation error when trying to create the wasm binary for file: assembly/__tests__/example.spec.ts.
Error: Runtime 'full' not found.
```

Looking at the docs about (runtimes)[https://www.assemblyscript.org/runtime.html] and the new [garbage collection implementation](https://www.assemblyscript.org/garbage-collection.html#runtime-interface) appears that the acceptable values for the `--runtime` flag in `as-pect.config.js` should now be "incremental", "minimal", and "stub".

However, it seems that there may be additional updates required, because when I try any of those values, I get the error:

```
$ npx  asp --verbose
       ___   _____                       __    
      /   | / ___/      ____  ___  _____/ /_   
[...]

PR #329, opened 2021-02-18
Update as-pect.config.js

bcolloran:
use  "incremental" for runtime value and change comment to mention acceptable AssemblyScript 0.18 runtime values

I'm not sure if this PR is what you want (if you have other changes you'd like to make at the same time, if "none" is still allowed, etc), but I thought I'd go ahead and offer this in case it's helpful and takes a little chore off your plate. :-) Feel free to ignore if this is not what you need

(see https://github.com/jtenner/as-pect/issues/328 )

### DIFF
--- a/packages/cli/init/as-pect.config.js
+++ b/packages/cli/init/as-pect.config.js
@@ -14,7 +14,7 @@ module.exports = {
     /** To output a wat file, uncomment the following line. */
     // "--textFile": ["output.wat"],
     /** A runtime must be provided here. */
-    "--runtime": ["full"], // Acceptable values are: full, half, stub (arena), and none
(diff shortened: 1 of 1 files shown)


willemneal (approved):
LGTM!  Thanks for your help!

jtenner:
@willemneal all my local testing, I've just removed that flag every time. Default is `"incremental"` and we should be encouraging people to use `asconfig.json` now. Perhaps we should deprecate the use of these flags altogether?

jtenner:
In the interest of not making breaking changes, I am very much in favor of pulling this request as-is. Perhaps down the road, we can provide this idea of a change when the compiler breaks something else.

PR #342, opened 2021-04-28
chore: switch to yarn workspaces

willemneal:
Since lerna supports yarn workspaces this should speed up the initialization of the project, though adding a dependency of using yarn.


willemneal:
Well I'm stumped why this isn't working.  @jtenner whenever you get a chance, could you test this out locally?

jtenner:
Looks like the ci bootstrapping process doesn't link the modules correctly. Lame.

jtenner:
Perhaps using yarn might be more effort than worth.

jtenner:
Let's merge @MaxGraey's pull request first, then try this again.

PR #356, opened 2021-06-09
Support for .as extension

munrocket:
Fixes: #355
This works locally for me but I don't know how to build and test packages


packages/cli/src/run.ts
```diff
@@ -350,8 +350,31 @@ export function run(cliOptions: Options, compilerArgs: string[]): void {
...
+    const dirs = [
+      path.resolve(__dirname, "../../assembly/assembly"),
+      path.resolve(__dirname, "../../assembly/assembly/internal"),
+      path.resolve(__dirname, "../../assembly/assembly/internal/comparison")
+    ];
+    dirs.forEach(dir => {
+      const files = fs.readdirSync(dir);
+      files.forEach(file => {
+        if (path.extname(file) === ".ts") {
+          const filename = file.split('.').slice(0, -1).join('.');
```
jtenner, line 365: I might refactor this to use the path library. You can use path.extname() and path.join(dir, basename + ext).

I'm also curious what happens if the user passes ".as" vs "as" what happens.

Otherwise this change seems reasonable.

jtenner (changes requested):
Some comments about obtaining the entry path using the node path library as personal preference and I had a question about how the extension flag works.

packages/cli/src/run.ts
munsocket: I am checked locally: ".as" and "as" is possible.
It's how AS compiler works right now https://github.com/AssemblyScript/assemblyscript/pull/1231/files

jtenner:
The ci issues here are not a problem. I should address the issue you uncovered at some point.

In the meantime, please update your snapshots by running the provided npm script in the root directory.

packages/cli/src/run.ts
```diff
@@ -350,8 +350,31 @@ export function run(cliOptions: Options, compilerArgs: string[]): void {
...
+          const newpath = `${dir}/${filename}.${newExt}`;
+          if (!fs.existsSync(newpath)) {
+            fs.copyFile(`${dir}/${file}`, newpath, err => console.log(err));
```
munsocket, line 368: Should be `copyFileSync` here I will fix it

munsocket:
> please update your snapshots by running the provided npm script in the root directory.

Doesn't see how. Updated only packages, but seems snapshots the same.
Test passed locally also with .as in my simple project works.

munsocket:
Tried `postinstall` and `test:update-snapshots` but nothing happens too much, because snapshots in .gitignore.
Is there something that I miss in PR? I am not used full functional of your test framework, but this solution works for me.

jtenner:
I can pull this fork and make changes today

munsocket:
Feel free to pull this in any time you want. I am just curious maybe I missed something. For example assembly/types or assembly/assembly/__tests__.

jtenner:
Hey don't worry. I'll take a look soon.

https://github.com/jtenner/as-pect/pull/356/checks?check_run_id=2802863853#step:5:904

As you can see here, the pointer changed on these references and that should not invalidate the snapshot. This is not your fault. I will take a look at this soon.

jtenner:
Do I have permission to push changes to your branch?

munrocket:
Added you as collaborator into my fork. Without this process you able only to create PRs.

jtenner:
I will swing around to this soon. Don't worry.

jtenner:
@munrocket this is not going to pass because of code coverage issues. This is unavoidable currently.

```ts
  // If another extension is used create copy of assembly/**/*.ts files
  if (flags["--extension"]) {
    console.log(chalk`{bgWhite.black [Log]} Changing extension for injected assembly files`);
    const newExt = flags["--extension"][0].replace(".", ""); // without dot should work
    const assemblyFolder = path.dirname(require.resolve("@as-pect/assembly/assembly/index.ts"));

    const files = glob.sync(path.join(assemblyFolder, "**/*.ts"));

    files.forEach(file => {
      const dirname = path.dirname(file);
      const basename = path.basename(file, path.extname(file));
      fs.copyFileSync(file, path.join(dirname, basename + "." + newExt));
    });
  }
  ```
  
  This does this trick, and looks a lot simpler.

**Comment by jtenner** [comment]
```
[Log] Effective command line args:
  [TestFile.ts] ..\as-pect\packages\assembly\assembly\index.as --runtime incremental --extension as --debug --binaryFile output.wasm --explicitStart --use ASC_RTRACE=1 --exportTable --importMemory --transform C:\Users\Joshua Tenner\projects\as-pect\packages\cli\node_modules\@as-pect\core\lib\transform\index.js     

✔ assembly/index.spec.as Pass: 1 / 1 Todo: 0 Time: 4.854ms
  [Result]: ✔ PASS
   [Files]: 1 total
  [Groups]: 2 count, 2 pass
   [Tests]: 1 pass, 0 fail, 1 total
    [Time]: 5279.49ms
```

munrocket:
awesome!

PR #359, opened 2021-06-16
add coverage... now to fix the bugs

jtenner:
This is going to be a very big deal. Currently, there is some issue with type information being carried over with `&&` and `||` operators and expressions that compile to a constant.

Example: `a && b` becomes `a && (__cover(num), b)`  which means that the right side of the expression is no longer constant.

I have refactored some of the code to help with this, but I have found some other side effect issues that I will need time to debug.

Any help would be appreciated.


jtenner:
Probably one of the most monumental parts of aspect, and I can't test it because coveralls isn't working well. 👀 

In any case, it works locally, and I am happy to support any bugs that arise.

PR #379, opened 2022-01-13
Fix toHaveLength assertion to accept nullable T

jtenner:
- [x] Update packages
- [x] Fix length assertion
- [x] Fix typescript error
- [ ] add coverage for reflect?
 
Looks like we have a few codepoints to add and fix...

```
│ assembly/internal/Reflect.ts                             │ 96.8% │ 98.2% │ 100% │ 93.3% │ 475:7, 489:46, 651:25 │
```


jtenner:
On `Reflect.ts` line 651, I believe it might be impossible to have a nullable `T` at this point.

willemneal:
Thanks!

PR #382, opened 2022-02-21
Watch Mode

mrchantey:
Here's an initial run at watch mode. Its working well on my project, but there are probably some changes needed before its ready to merge, a few things in particular:
- Kill Process
  - We probably want to ensure that the previous run gets cleaned up properly. Maybe somebody more familiar with the codebase has an idea of the best way to do this.
- Use config for the watch pattern
  - at the moment I'm using '**/*.ts' which is enough to get started, but probably not suitable for all codebases.
  - I suppose it involves moving some functions outside of `run` so `watch` can use them as well
- Chokidar depencency / package.lock
  - Landed on `chokidar` because it looks like thats what jest does, but there may be a better solution
  - I'm not sure if the package-lock behaved itself when I installed chokidar, looks like a lot of changes even aside from the whitespace ones.

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -1,867 +1,1380 @@
 {
-	"name": "@as-pect/cli",
-	"version": "6.2.4",
-	"lockfileVersion": 2,
-	"requires": true,
-	"packages": {
-		"": {
-			"name": "@as-pect/cli",
-			"version": "6.2.4",
-			"license": "MIT",
-			"dependencies": {
-				"@as-covers/core": "0.2.1",
-				"chalk": "^4.1.1",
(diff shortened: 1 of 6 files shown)


mrchantey:
yoooooo speaking of 'blazing fast', i noticed that this watch mode was much faster than using a child-process method like `onchange`. I think that V8 is performing JIT operations every time it runs:

Using `onchange` and for first time in `watch mode` it takes about half a second:

Screenshot 2022-02-21 135714

then on a file change in watch mode:
Screenshot 2022-02-21 135817

on second file change:
Screenshot 2022-02-21 135834

on nth file change:
Screenshot 2022-02-21 135904

PR #392, opened 2022-08-27
Get further

trusktr:
Main changes:

- remove @as-pect/assemblyscript
- list @as-pect/transform as a devDependency so it is linked in (this was the problem)
- a couple updates in package.jsons to abide by ESM rules (`npm test` in transform/ didn't work)
- the rest is just formatting


packages/cli/package.json
```diff
@@ -2,7 +2,10 @@
   "name": "@as-pect/cli",
   "version": "6.2.4",
   "description": "Write AssemblyScript tests at blazing fast speeds",
-  "main": "lib/index.js",
+  "main-OFF": "lib/index.js",
```
trusktr, line 5: out with the old, in with the new!

packages/transform/src/createAddReflectedValueKeyValuePairsMember.ts
```diff
@@ -177,10 +161,7 @@ function createIsDefinedIfStatement(
         // super.__aspectAddReflectedValueKeyValuePairs
         TypeNode.createPropertyAccessExpression(
           TypeNode.createSuperExpression(range),
-          TypeNode.createIdentifierExpression(
-            "__aspectAddReflectedValueKeyValuePairs",
-            range,
-          ),
+          TypeNode.createIdentifierExpression("__aspectAddReflectedValueKeyValuePairs", range),
```
trusktr, line 164: Mostly a bunch of formatting from printWidth 120. Makes the code not so tall.

PR #397, opened 2022-11-13
Upgrade to as23

jtenner:
It's worth noting that upgrading to the latest compiler with constant breaking changes requires some fortitude.

This PR will be squashed and merged.


jtenner:
This ended up being a great pull request, and it is now compatible with as 24. The reason why is because things need to be downcast as object in order to be tested properly for strict equality. This pull request was an absolute journey.

PR #400, opened 2022-11-14
Fix CI

CountBleck:
This PR should hopefully fix CI. Right now, it's just me poking around to see what needs to be fixed.


jtenner:
Can we update to latest node version too?

PR #402, opened 2022-12-07
move @assemblyscript/loader to core dependencie

peterjah:


jtenner:
Looks like coveralls is broken. Definitely not a problem for this pull request. LGTM.

